PR #6, opened 2019-11-26
update sdk for windows builds

dovholuknf:
* don't include the 'm' library if not on linux (builds fine with windows also see https://stackoverflow.com/questions/54935559/linking-math-library-in-cmake-file-on-windows-and-linux)
* add the dll project back when on windows
* define strncasecmp  for windows


library/model.c
```diff
@@ -14,14 +14,12 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+#include <time.h>
 
 #if _WIN32
 #define timegm(v) _mkgmtime(v)
 #else
 #define _GNU_SOURCE
```
ekoby: If I remember correctly, _GNU_SOURCE has to be defined before #include <time.h> otherwise it has no effect

dovholuknf: i've added the include in both branches of the if to address this. i chose to do that instead of adding another if block. figured it was marginally better than another if...

library/model.c
```diff
@@ -14,14 +14,12 @@ See the License for the specific language governing permissions and
...
 
-
 #if _WIN32
+#include <time.h>
 #define timegm(v) _mkgmtime(v)
 #else
-#define _GNU_SOURCE
-
+#define _GNU_SOURCE //add time.h include after defining _GNU_SOURCE
 #include <time.h>
```
dovholuknf: added time.h after define of _GNU_SOURCE per PR comment

PR #17, opened 2020-01-06
update copyrights to be in sync with 2020

dovholuknf:


tests/test_model.cpp
```diff
@@ -18,13 +18,13 @@ limitations under the License.
...
 
 #if _WIN32
-#include <windows.h>
-#define timegm(v) _mkgmtime(v)
-#define gmtime(v) _gmtime32(v)
-#else
-#define _GNU_SOURCE //add time.h include after defining _GNU_SOURCE
-#include <time.h>
-#endif
+#include <windows.h>
```
dovholuknf: couple of files got a dos2unix applied. so perhaps change to ignore whitespace

PR #20, opened 2020-01-07
Better controller comm 15

### LINKED ISSUE #15: improve controller communication
request to controller should be made on the event loop instead of background thread
enable periodic polling of available services

ekoby:
Closes #15 

this changes SDK-to-EdgeController communication to be fully async and executed on the event loop provided to SDK.

Services and Network Sessions are no longer preloaded during NF_init() but instead created on demand.


CMakeLists.txt
```diff
@@ -20,7 +20,8 @@ if (WIN32)
     message("WIN32 build. Creating: ${CMAKE_BINARY_DIR}/cmake_install/ziti-sdk-${PROJECT_VERSION}")
     file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/cmake_install/ziti-sdk-${PROJECT_VERSION})
     set(CMAKE_INSTALL_PREFIX ${CMAKE_BINARY_DIR}/cmake_install/ziti-sdk-${PROJECT_VERSION})
-    add_subdirectory(windows)
+    # TODO 
+    # add_subdirectory(windows)
```
dovholuknf: perhaps make an issue for this?

ekoby: I added general issue for build a shared lib already #24

inc_internal/ziti_ctrl.h
```diff
@@ -0,0 +1,57 @@
+/*
+Copyright (c) 2019 Netfoundry, Inc.
```
dovholuknf: moving a file is a 'touch'? needs 2019-2020?

ekoby: missed it in update. fixed

library/controller.h
```diff
@@ -24,7 +24,7 @@ typedef int (*ctrl_req)(struct nf_ctx *, uv_os_sock_t sock, tls_engine *ssl);
 int ziti_ctrl_process(nf_context ctx, ...);
 
 int ziti_ctrl_version(nf_context ctx, uv_os_sock_t, tls_engine *);
-int ziti_ctrl_login(nf_context ctx, uv_os_sock_t, tls_engine *);
+int ziti_ctrl_login_old(nf_context ctx, uv_os_sock_t, tls_engine *);
```
dovholuknf: needed?

dovholuknf:
There's a mix of free vs FREE used - do you want to use FREE everywhere?

inc_internal/ziti_ctrl.h
```diff
@@ -0,0 +1,57 @@
+/*
+Copyright (c) 2019-2020Netfoundry, Inc.
```
dovholuknf: almost: 2020Netfoundry

library/ziti.c
```diff
@@ -307,3 +318,33 @@ extern int NF_listen(nf_connection serv_conn, const char *service, nf_listen_cb
...
+    if (err != NULL) {
+        ZITI_LOG(ERROR, "failed to get controller version from %s:%s %s(%s)",
+                 ctrl->client.host, ctrl->client.port, err->code, err->message);
+        free_ziti_error(err);
+    }
+    else {
+        ZITI_LOG(INFO, "connected to controller %s:%s version %s(%s %s)",
+                 ctrl->client.host, ctrl->client.port, v->version, v->revision, v->build_date);
+        free_ctrl_version(v);
+    }
```
dovholuknf: you aren't free'ing the err here like in other functions - intentional?

dovholuknf: lol - the fucntion above does it - just didn't notice somehow!

library/ziti_ctrl.c
```diff
@@ -0,0 +1,293 @@
...
+        }
+    }
+}
+
+static void ctrl_default_cb (void *s, ziti_error *e, struct ctrl_resp *resp) {
+    if (resp->resp_cb) {
+        resp->resp_cb(s, e, resp->ctx);
+    }
+
+    free(resp);
```
dovholuknf: what will free the resp->ctx and resp->body etc? that all handled elsewhere?

library/ziti_ctrl.c
```diff
@@ -0,0 +1,293 @@
...
+
+static void ctrl_login_cb(ziti_session *s, ziti_error *e, struct ctrl_resp *resp) {
+    resp->ctrl->session = strdup(s->token);
+    um_http_header(&resp->ctrl->client, "zt-session", s->token);
+    ctrl_default_cb(s, e, resp);
+}
+
+static void ctrl_logout_cb(void *s, ziti_error *e, struct ctrl_resp *resp) {
+    FREE(resp->ctrl->session);
+    um_http_header(&resp->ctrl->client, "zt-session", NULL);
```
dovholuknf: there's no way to just remove a header? or is setting a header to NULL actually removing it from the client?

library/ziti_ctrl.c
```diff
@@ -0,0 +1,293 @@
...
+    ctrl_default_cb(s, e, resp);
+}
+
+static void ctrl_service_cb(ziti_service **services, ziti_error *e, struct ctrl_resp *resp) {
+    ziti_service *s = services != NULL ? services[0] : NULL;
```
dovholuknf: if services isn't null there's guaranteed to be at least one service? if all services are removed from the identity at the controller will services get set to NULL?

library/ziti_ctrl.c
```diff
@@ -0,0 +1,293 @@
...
+        resp->received += len;
+    } else if (len == UV_EOF) {
+        const char* data;
+        int data_len;
+        void *resp_obj = NULL;
+        ziti_error *err = NULL;
+
+        if (resp->status > 299) {
+            mjson_find(resp->body, resp->received, "$.error", (const char **) &data, &data_len);
+            err = parse_ziti_error(data, data_len);
```
dovholuknf: 300's are 'errors'?

library/ziti_ctrl.c
```diff
@@ -0,0 +1,293 @@
...
+    resp->ctrl_cb = ctrl_service_cb;
+
+    req->data = resp;
+}
+
+void ziti_ctrl_get_net_session(
+        ziti_controller *ctrl, ziti_service *service,
+        void (*cb)(ziti_net_session *, ziti_error*, void*), void* ctx) {
+
+    char *content = malloc(128);
```
dovholuknf: anything free'ing content?

library/ziti_ctrl.c
```diff
@@ -0,0 +1,293 @@
...
+    size_t len = (size_t) sprintf(content, "{\"serviceId\":\"%s\",\"hosting\":%s}",
+                                      service->id, service->hostable ? "true" : "false");
+
+    um_http_req_t *req = um_http_req(&ctrl->client, "POST", "/sessions");
+    req->resp_cb = ctrl_resp_cb;
+    req->body_cb = ctrl_body_cb;
+    um_http_req_header(req, "Content-Type", "application/json");
+    um_http_req_data(req, content, len, free_body_cb);
```
dovholuknf: can len be > 128?

library/controller.h
ekoby: file removed

inc_internal/ziti_ctrl.h
ekoby: yep

library/ziti_ctrl.c
ekoby: setting to NULL === remove

library/ziti_ctrl.c
ekoby: we don't expect them since controller does not do redirects

library/ziti_ctrl.c
ekoby: services is  NULL terminated array, if we get empty array then services[0] would be NULL

library/ziti_ctrl.c
ekoby: we're formatting request above, so why would it be >128?

library/ziti_ctrl.c
ekoby: `resp->ctx` is client owned data, `resp->body` is freed after it's parsed into an API object. line 130

library/ziti_ctrl.c
ekoby: line 266, freed in the callback

dovholuknf: Ah, content == body... Thanks

PR #26, opened 2020-01-09
handle login errors, set proper content-type

ekoby:


library/ziti_ctrl.c
```diff
@@ -84,8 +84,14 @@ static void ctrl_default_cb (void *s, ziti_error *e, struct ctrl_resp *resp) {
...
+    if (e) {
+        ZITI_LOG(ERROR, "%s(%s)", e->code, e->message);
+    }
+
+    if (s) {
+        resp->ctrl->session = strdup(s->token);
+        um_http_header(&resp->ctrl->client, "zt-session", s->token);
+    }
     ctrl_default_cb(s, e, resp);
 }
```
dovholuknf: should this be calling free_ziti_err ?

ekoby: it is freed by the client code (in client callback)

PR #28, opened 2020-01-10
minor fixes

ekoby:


library/ziti.c
```diff
@@ -325,13 +335,18 @@ static void session_cb(ziti_session *session, ziti_error *err, void *ctx) {
     int errCode = err ? code_to_error(err->code) : ZITI_OK;
 
     if (init_req != NULL) {
-        init_req->nf->session = session;
+        if (session) {
+            init_req->nf->session = session;
+            ZITI_LOG(INFO, "logged in successfully => api_session[%s]", session->id);
```
dovholuknf: should session id be logged at all? seems like maybe a security leak?

library/ziti_ctrl.c
```diff
@@ -26,6 +26,12 @@ limitations under the License.
 #define MJSON_API_ONLY
 #include <mjson.h>
 
+#if _WIN32
+
+#define strcasecmp _stricmp
```
dovholuknf: HURRAY!

library/ziti.c
ekoby: token is the key, can't use id to login

PR #31, opened 2020-01-14
Fix.eol

andrewpmartinez:


.gitattributes
```diff
@@ -0,0 +1,42 @@
...
+*.ps1 text eol=lf
+*.g4 text eol=lf
+*.interp text eol=lf
+*.pem text eol=lf
+*.cnf text eol=lf
+*.conf text eol=lf
+*.gitmodules text eol=lf
+*.variants text eol=lf
+*.cmake text eol=lf
+*.bat text eol=lf
```
scareything: should .bat (and .cmd) use crlf?

dovholuknf: had the same thought... i was in the midst of testing to make sure it'll still work when andrew merged. it seems like .bat at least doesn't care so it seems fine

dovholuknf: uh - he musta merged something else cause... i thought for sure this one was merged but it clearly says 'open'... so... yeah i'm just going blind it seems :(

PR #32, opened 2020-01-14
set uv_mbed logging, some cleanup

ekoby:


library/ziti.c
```diff
@@ -207,6 +208,8 @@ int NF_set_timeout(nf_context ctx, int timeout) {
 int NF_shutdown(nf_context ctx) {
```
dovholuknf: i feel like the freeing of the context session as part of this should be added to the header file doc, no? should the whole nf_context get free'ed here too? maybe ziti_close_channels does that?

ekoby: I am not sure. I think lifecycle is better discussed/explained in user guide. API doc is probably not the right place for it.

PR #42, opened 2020-01-20
Doxygen updates 2

dovholuknf:
Updates to provide API doc for the C sdk. These changes are available here https://dovholuknf.github.io/ziti-doc/api/clang/api/index.html for (hopefully) easier preview


CMakeLists.txt
```diff
@@ -12,6 +12,8 @@ message("   branch : ${GIT_BRANCH}")
 message("     hash : ${GIT_COMMIT_HASH}")
 message("     date : ${BUILD_DATE}")
 
+set(ZITI_HEADER_FILES ${PROJECT_SOURCE_DIR}/includes/nf/errors.h ${PROJECT_SOURCE_DIR}/includes/nf/ziti.h)
```
ekoby: this is no longer used, right?

dovholuknf: it's referenced in library/CMakeLists.txt but not in the example programs

ekoby: it should be moved to the one in library cmake then

dovholuknf: moved

PR #47, opened 2020-01-28
Bug fixes

ekoby:


dovholuknf:
can you merge in master too please to pick up the doc change?

includes/nf/errors.h
```diff
@@ -30,6 +30,7 @@ limitations under the License.
     XX(CONNABORT, "Connection to edge router terminated") \
     XX(KEY_SPEC_UNSUPPORTED, "unsupported key specification") \
     XX(KEY_INVALID, "invalid key") \
+    XX(INVALID_STATE, "invalid state") \
```
dovholuknf: can you add some doc for this similar to what's there now pls

library/connect.c
```diff
@@ -482,8 +523,16 @@ int ziti_accept(nf_connection conn, nf_conn_cb cb, nf_data_cb data_cb) {
...
+    LIST_INSERT_HEAD(&conn->nf_ctx->connect_requests, req, _next);
+
     ziti_channel_send_for_reply(ch, content_type, headers, 3, (const uint8_t *) &clt_conn_id, sizeof(clt_conn_id),
                                 connect_reply_cb, req);
 
     return ZITI_OK;
 }
+
+int ziti_process_connect_reqs(nf_context nf) {
+    ZITI_LOG(WARN, "TODO");
```
dovholuknf: maybe leave a link to an issue here instead?

library/ziti_ctrl.c
```diff
@@ -64,18 +64,27 @@ struct ctrl_resp {
...
-            break;
-        }
-        if (strcasecmp(h->name, "Transfer-Encoding") == 0 && strcmp(h->value, "chunked") == 0) {
-            resp->resp_chunked = true;
-            resp->body = malloc(0);
+    if (code < 0) {
+        NEWP(err, ziti_error);
+        err->code = strdup("CONTROLLER_UNAVAILABLE");
+        err->message = strdup(uv_strerror(code));
+        ctrl_default_cb(NULL, err, resp);
```
dovholuknf: i don't see ctrl_default_cb free'ing the err nor the err->code and err->message - is that a leak or is it cleaned up somewhere else?

includes/nf/errors.h
ekoby: done

library/connect.c
ekoby: this is not used yet. Maybe never will. I'll resolve one way or the other

library/ziti_ctrl.c
ekoby: it is up to the code that calls controller to keep track or free results and/or errors

PR #49, opened 2020-01-30
add the ability for the host example to be a sample client too

dovholuknf:
dual purpose the sample-host example so that it can be both 'server' and 'client'


ekoby:
I'd rather see sample-client as a separate executable. It would be easier to understand (in the context of quick-start).

dovholuknf:
i don't agree that it'd be easier to understand. i find it easier to just build one library and use it. i'll cite netcat as an example - you use it as both host and client and that's why i think it should stay as one example.

ekoby:
`netcat` is a general purpose tool, not a tutorial level exercise and it is not how majority of ziti use cases would work. IF we decide to keep one executable we should at least separate hosting and client code into separate files.

PR #58, opened 2020-03-04
end-to-end crypto

ekoby:


library/connect.c
```diff
@@ -370,11 +381,86 @@ int ziti_disconnect(struct nf_conn *conn) {
     return uv_async_send(ar);
 }
 
+static void crypto_wr_cb(nf_connection conn, ssize_t status, void* ctx) {
+
```
smilindave26: nuffin'?

ekoby: added error reporting and callbacking

PR #60, opened 2020-03-19
Idempotent shutdown

scareything:
Avoid crashing due to double-free, assertions, etc. when NF_shutdown is called twice.

### DIFF
--- a/library/channel.c
+++ b/library/channel.c
@@ -104,7 +104,7 @@ void ziti_close_cb(uv_handle_t *h) {
 
 int ziti_channel_close(ziti_channel_t *ch) {
     int r = 0;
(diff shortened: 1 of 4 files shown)


library/channel.c
```diff
@@ -103,7 +103,12 @@ void ziti_close_cb(uv_handle_t *h) {
 }
 
 int ziti_channel_close(ziti_channel_t *ch) {
-    return uv_mbed_close(&ch->connection, ziti_close_cb);
+    int r = 0;
+    if (ch->state == Connected) {
```
dovholuknf: one can't close a channel in a 'bad' state? should this be != closed instead?

PR #71, opened 2020-04-03
Service config

ekoby:
make `ziti_service` part of SDK API
improve `nf_service_cb` to include ziti_service object
add access to service configs


library/ziti.c
```diff
@@ -180,19 +184,61 @@ int NF_init(const char* config, uv_loop_t* loop, nf_init_cb init_cb, void* init_
...
+        return ZITI_INVALID_CONFIG;
+    }
+
     nf_config *cfg = NULL;
-    tls_context *tls = NULL;
+    if (opts->config != NULL) {
+        TRY(ziti, load_config(opts->config, &cfg));
+    }
+    if (opts->controller == NULL) {
+        opts->controller = strdup(cfg->controller_url);
```
dovholuknf, line 198: i don't see opts->controller getting free'ed in here? either i missed it or it's done somewhere else so i'm just commenting wondering if that's the case?

PR #77, opened 2020-04-16
elevate ZITI_LOG to API

ekoby:


dovholuknf:
probably should have some words for the api docs, right?

ekoby:
> probably should have some words for the api docs, right?

sure

PR #83, opened 2020-04-22
Win fixes

ekoby:
Centralize disposal of nf_connection
fix memory issues
win32 misc

### DIFF
--- a/library/connect.c
+++ b/library/connect.c
@@ -55,6 +55,20 @@ static void free_conn_req(struct nf_conn_req *r) {
     free(r);
 };
 
+int close_conn_internal(struct nf_conn *conn) {
+    if (conn->state == Closed && conn->write_reqs == 0) {
+        LIST_REMOVE(conn, next);
+        FREE(conn->rx);
+        conn->flusher->data = NULL;
+        uv_close((uv_handle_t *) conn->flusher, free_handle);
+        buffer_cleanup(conn->inbound);
+        free_buffer(conn->inbound);
+        FREE(conn);
+        return 1;
+    }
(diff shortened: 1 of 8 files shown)


library/ziti.c
```diff
@@ -571,3 +578,22 @@ static const ziti_version sdk_version = {
...
+static void grim_reaper(uv_prepare_t *p) {
+    nf_context nf = p->data;
+
+    int count = 0;
+    const char *url;
+    ziti_channel_t *ch;
+    MODEL_MAP_FOREACH(url, ch, nf->channels) {
+        nf_connection conn = LIST_FIRST(&ch->connections);
+        while(conn != NULL) {
+            nf_connection try_close = conn;
```
dovholuknf, line 591: this should have some kind of log associated with it shouldn't it? This will be a really hard bug to find with no trace of where the connection went if there's a bug?

library/connect.c
```diff
@@ -421,6 +427,10 @@ static int establish_crypto (nf_connection conn, message *msg) {
 
 static void flush_to_client(uv_async_t *fl) {
     nf_connection conn = fl->data;
+    if (conn == NULL || conn->state == Closed) {
```
dovholuknf, line 430: should this be after the log on line 424/434?

ekoby: nope -- can't check buffer if conn is NULL

dovholuknf: How about a different message then in this if?

library/ziti.c
ekoby: added log msg in conn_close_internal

PR #89, opened 2020-04-30
Metrics phase1: initial transfer rate

ekoby:


includes/nf/ziti.h
```diff
@@ -125,7 +125,7 @@ typedef struct nf_options_s {
     nf_init_cb init_cb;
     nf_service_cb service_cb;
 
-    int refresh_interval;
+    long refresh_interval;
```
dovholuknf, line 128: what made you change this into a long? is this in millis?

ekoby: no, it just cleared a warning

PR #93, opened 2020-05-04
Cleanup markdown for better viewing on github

smilindave26:
Things like `[TOC]` and `/subsection` were bleeding through...

### DIFF
--- a/building.md
+++ b/building.md
@@ -1,23 +1,21 @@
-\page building Building the Project
+# Building the Project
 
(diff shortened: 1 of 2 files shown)


building.md
```diff
@@ -1,23 +1,21 @@
-\page building Building the Project
```
dovholuknf, line 1: these are ugly sure but they are used by doxygen. i've not found a way to have our cake and eat it too here... we either have doxygen doc look nice - or github.

PR #96, opened 2020-05-10
added NF_enroll_with_key

smilindave26:
Added NF_enroll_with_key to let me bring my own private key to the enrollment process

### DIFF
--- a/library/ziti_enroll.c
+++ b/library/ziti_enroll.c
@@ -156,6 +156,10 @@ int verify_es512(struct enroll_cfg_s *ecfg, mbedtls_pk_context *ctx) {
 }
 
 int NF_enroll(const char* jwt_file, uv_loop_t* loop, nf_enroll_cb external_enroll_cb, void* external_enroll_ctx) {
+    return NF_enroll_with_key(jwt_file, NULL, loop, external_enroll_cb, external_enroll_ctx);
+}
+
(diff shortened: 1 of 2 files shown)


library/ziti_enroll.c
```diff
@@ -156,6 +156,10 @@ int verify_es512(struct enroll_cfg_s *ecfg, mbedtls_pk_context *ctx) {
 }
 
 int NF_enroll(const char* jwt_file, uv_loop_t* loop, nf_enroll_cb external_enroll_cb, void* external_enroll_ctx) {
+    return NF_enroll_with_key(jwt_file, NULL, loop, external_enroll_cb, external_enroll_ctx);
```
ekoby, line 159: I'd mode keygen step here, and require `NF_enroll_with_key()` to be called with non-NULL key

smilindave26: yes, that's better.  Thinking about refactoring the enrollment a bit.
* add tls_context as well as key here (pass default_tls_context() into NF_enroll_with_key() from NF_enroll()
* remove inline tls handshake in NF_enroll() that is used to get server's public key, and instead grab servers key in the callback from getting well-known certs (so in that callback, validates the JWT, store off the CA pool, and call `/enroll`  
* Looks like will need to add API to TLS engine configure auth mode (VERIFY_NONE when getting the CA poll, VERIFY_REQUIRED for the enrollment.

That refactor is somewhat significant.  I might do ALAP and just switch the tls_context for the /enroll call for this PR and leave the initial handshake / JWT verification as is with a follow-up PR for the larger change

PR #109, opened 2020-06-25
5s moving avg

dovholuknf:
* moves rate_type enum into a new enums.h.
* exposes the rate type choice to someone looking to influence it via ziti_options_s


library/metrics.c
```diff
@@ -62,9 +62,13 @@ extern void metrics_rate_close(rate_t* r) {
     LIST_REMOVE(r, _next);
 }
 
-extern void metrics_rate_init(rate_t *r, enum rate_type type) {
+extern void metrics_rate_init(rate_t *r, rate_type type) {
     memset(r, 0, sizeof(rate_t));
     switch (type) {
+        case EWMA_5s:
+            r->tick_fn = tick_ewma;
+            *(double*)(&r->param) = 1.0 - pow(M_E, -(interval * 6.0)); //for shorter durations like this - weight it more
```
ekoby, line 70: this formula does not seem right for 5s average
comparing to other ones I expect `1.0 - pow(M_E, -1)`

dovholuknf: That's indeed what I had originally but in practice it doesn't behave/feel right. This function is the "scaling tractor" and from the research I did it's entirely related to "taste". It's a "weighting factor". If you want to discuss it live let me know.

ekoby: with the above parameter the previous value is almost completely discarded. so we should call it something else -- instant rate, or something like that, and have a separate calculation accordingly

dovholuknf:
declining in favor of 'instant' metrics: https://github.com/openziti/ziti-sdk-c/pull/120

PR #120, opened 2020-07-06
Instant metrics

dovholuknf:
Add a way to get the 'latest' metrics: INSTANT

* extract the enum to a .h and include it where needed
* keep EWMA_5s but put the param calculation to be congruent to the other formulas
* remove redundancy from EWMA_1m param function
* convert some WARN msgs to DEBUG in proxy-c


programs/ziti-prox-c/proxy.c
```diff
@@ -385,6 +385,7 @@ void run(int argc, char **argv) {
             .refresh_interval = 600,
             .ctx = &listeners,
             .config_types = my_configs,
+            .metrics_type = INSTANT,
```
dovholuknf, line 388: i left this here as illustration on how to change the metric type

inc_internal/metrics.h
```diff
@@ -40,28 +41,22 @@ struct rate_s {
     atomic_llong param;
 
     void (*tick_fn)(struct rate_s *);
+    double (*rate_get_fn)(struct rate_s *); //a function that will return the rate as a double
```
ekoby, line 44: i don't think you need this func:
* just use `instant_rate()` in `tick_instant` and store it
* regular `metric_get_rate()` calc would work

dovholuknf: `instant_rate` divides by nanos not seconds. tick_instant does the division ahead of time. i could multiply it by nanos in `tick_instant` if that's what you want? i could then remove the functions stuff i added. this way seemed better than that but i don't care terribly if you want me to rm the functions

dovholuknf: looks like i could set intervalNanos to 5 - but that also felt awkward. i don't see it used anywhere else though so maybe i'll jsut rename that variable and use it

dovholuknf: i tried it out... metrics_rate_get also multiplies by `SECOND` which causes another, different wrinkle - i think the way i did it here is the most clear so far

ekoby: I pretty sure you'd get the same result: interval is 5 seconds, intervalNanos is interval in nanos, metric_get_rate scales is it up to rate-per-second. metric->rate would be stored in value/ns and consistent with the way other rates are handled

ekoby: > looks like i could set intervalNanos to 5 - but that also felt awkward. i don't see it used anywhere else though so maybe i'll jsut rename that variable and use it

`intervalNanos` is used in `instant_rate()`

dovholuknf: i've removed the functions at your request

library/metrics.c
```diff
@@ -150,3 +164,15 @@ static void tick_ewma(rate_t *ewma) {
         InterlockedExchange(&ewma->init, 1);
     }
 }
+
+// a function that will return the last measured value expressed as 
+double metrics_get_instant(rate_t* r) {
```
ekoby, line 169: this is not needed, right?

dovholuknf: removed

library/ziti.c
```diff
@@ -571,17 +571,25 @@ static void session_cb(ziti_session *session, ziti_error *err, void *ctx) {
         }
 
         if (ztx->opts->refresh_interval > 0 && !uv_is_active((const uv_handle_t *) &ztx->refresh_timer)) {
+            ZITI_LOG(INFO, "refresh_interval set to [%d] ms", ztx->opts->refresh_interval * 1000);
```
ekoby, line 574: any reason log in ms instead of seconds?

library/metrics.c
```diff
@@ -43,6 +43,10 @@ static LIST_HEAD(meters, rate_s) all_rates = LIST_HEAD_INITIALIZER(all_rates);
 static void ticker_cb(uv_timer_t *t);
 static void tick_ewma(rate_t *ewma);
 static void tick_cma(rate_t *cma);
+static void tick_instant(rate_t *inst);
+
+static double default_rate_get(rate_t* r);
+static double metrics_get_instant(rate_t *r);
```
ekoby, line 50: obsolete declarations

library/ziti.c
dovholuknf: hunh. nope not really sure what happened there. i think i might have misinterpreted the variable at one point - i'll conver that over to seconds

dovholuknf: oh it's probably because of the next line: ```
            uv_timer_start(&ztx->refresh_timer, services_refresh, 0, ztx->opts->refresh_interval * 1000);```

dovholuknf: i changed it to seconds

PR #123, opened 2020-07-09
Linux embedded build

ekoby:


scareything:
fyi cmake fails for me on Darwin when I do it from an empty directory. It works if I run it again. It feels vaguely familiar somehow... I'm taking a look.

PR #124, opened 2020-07-10
re-auth if SDK gets UNAUTHORIZED from controller

ekoby:


library/ziti.c
```diff
@@ -579,7 +595,16 @@ static void session_cb(ziti_session *session, ziti_error *err, void *ctx) {
         }
 
     } else {
-        ZITI_LOG(ERROR, "failed to login: %s[%d](%s)", err->code, errCode, err->message);
+        if (ztx->session) {
+            ZITI_LOG(WARN, "failed to refresh: %s[%d]", err->code, errCode);
+            if (strcmp(err->code, "UNAUTHORIZED") == 0) {
+                ziti_re_auth(ztx);
```
smilindave26, line 601: What happens when an identity is deleted at controller but sdk still active (eg identity still exists at the Tunneler)?

ekoby: > What happens when an identity is deleted at controller but sdk still active (eg identity still exists at the Tunneler)?

the first time SDK gets `UNAUTHORIZED` the `ztx->session` is cleared and SDK tries to login again, fails and after that all background refresh is stopped.

I guess I need to add notification that all services are no longer available, is that your thinking?

smilindave26: yeah, that would help.  And if the auth issue was somehow 'fixed', telling the user to restart their tunneler seems reasonable (versus doing something like exponential backoff retires)

ekoby: added more cleanup code:
* on api-session de-auth -- cleanup sessions
* on failure to login -- notify app that services are not available

PR #126, opened 2020-07-13
improve conn buffer cleanup

ekoby:
fix buffer leak of undelivered data
better handle errors reported by ziti_data_cb

### DIFF
--- a/library/buffer.c
+++ b/library/buffer.c
@@ -52,6 +52,12 @@ buffer *new_buffer() {
 }
 
 void free_buffer(buffer* b) {
+    while (!STAILQ_EMPTY(&b->chunks)) {
+        chunk_t *chunk = STAILQ_FIRST(&b->chunks);
+        STAILQ_REMOVE_HEAD(&b->chunks, next);
+        free(chunk->buf);
+        free(chunk);
+    }
     free(b);
 }
 
--- a/library/connect.c
+++ b/library/connect.c
@@ -64,7 +64,10 @@ int close_conn_internal(struct ziti_conn *conn) {
         FREE(conn->rx);
         conn->flusher->data = NULL;
         uv_close((uv_handle_t *) conn->flusher, free_handle);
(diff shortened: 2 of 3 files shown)


library/buffer.c
```diff
@@ -52,6 +52,12 @@ buffer *new_buffer() {
 }
 
 void free_buffer(buffer* b) {
+    while (!STAILQ_EMPTY(&b->chunks)) {
+        chunk_t *chunk = STAILQ_FIRST(&b->chunks);
+        STAILQ_REMOVE_HEAD(&b->chunks, next);
+        free(chunk->buf);
```
scareything, line 58: Is it worth using FREE here, incase someone is holding a reference to `chunk`?

library/connect.c
```diff
@@ -64,7 +64,10 @@ int close_conn_internal(struct ziti_conn *conn) {
         FREE(conn->rx);
         conn->flusher->data = NULL;
         uv_close((uv_handle_t *) conn->flusher, free_handle);
-        buffer_cleanup(conn->inbound);
+        if (buffer_available(conn->inbound) > 0) {
+            ZITI_LOG(WARN, "dumping %zd bytes of undelivered data conn[%d]",
```
dovholuknf, line 68: it might be useful to log that the connection is closing in this log msg. line 62 (above) will only log if verbose is on and that seems like relevant information to someone running at warn level.

`closing connection - dumping bytes....`

library/buffer.c
ekoby: that wouldn't help, chunk is a local var

library/connect.c
ekoby: that's implied, no? this message is only logged when connection is closed

dovholuknf: i would have no clue looking through a log that this is only logged when the connection is closed/closing

ekoby: sure you would. function name(`close_conn_internal()`) is logged as well

PR #132, opened 2020-07-30
update uv-mbed

rentallect:


library/ziti.c
```diff
@@ -390,6 +390,9 @@ int ziti_write(ziti_connection conn, uint8_t *data, size_t length, ziti_write_cb
     req->cb = write_cb;
     req->ctx = write_ctx;
 
+    ZITI_LOG(TRACE, "conn is: %p", conn);
```
dovholuknf, line 393: @ekoby this just makes me wonder if there is any way to #def out trace calls? or perhaps neuter it so it's compiled out somehow and trace logging needs to be enabled? or maybe a second function that's specifically for 'dev type debugging' that is #def'ed in or somesuch? logging in hotpaths like this always makes me wonder

rentallect: I like where your head is at, but we should do that work in a separate task.  If you are bothered in the short term by the slight overhead this adds in our current impl, I am happy to simply purge many of the TRACE's I sprinkled about as we were chasing down the uv-mbed defect.

dovholuknf: yeah it definitely held me back from clicking the approve button immediately. i'd like @ekoby to weigh in here. If it were me I would probably make a ZITI_DEBUG function that is #def'ed in/out - you could stub that function out here and just delegate it straigt to ZITI_LOG (same params etc) so that it doesn't slow you down or anything

ekoby: ZITI_LOG is a noop (1 comparison) is level is not enabled

library/connect.c
```diff
@@ -95,6 +98,7 @@ void on_write_completed(struct ziti_conn *conn, struct ziti_write_req_s *req, in
 
         if (status < 0) {
             conn->state = Closed;
+            ZITI_LOG(TRACE, "conn %p state is now Closed", conn);
```
ekoby, line 101: I'd prefer to add `conn_id` here and other places where we log with `%p`

rentallect: done

PR #141, opened 2020-08-31
add a helper function for reuse to return the format of the log

dovholuknf:


library/utils.c
```diff
@@ -201,3 +204,16 @@ void hexDump (char *desc, void *addr, int len) {
...
     ZITI_LOG(DEBUG, " ");
 }
+
+void ziti_log_format(char* time_str, size_t time_str_len, uv_timeval64_t start_time) {
+#if _WIN32
+    time_t rawtime = start_time.tv_sec;
+    struct tm  start_tm;
+
+    start_tm = *localtime(&rawtime);
+    strftime(time_str, time_str_len, "%Y-%m-%dT%H:%M:%S", &start_tm);
```
dovholuknf, line 214: @rentallect - i pust `strftime` back but with an updated formatter. the `%FT%T` format crashed on that line for me as well but with the different format string it doesn't crash. looking over at https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strftime-wcsftime-strftime-l-wcsftime-l?view=vs-2019 - I will try using `%F%T` which seems to be the equivalent of `%FT%T` in microsoft land

dovholuknf: after refactoring a bit - i was not able to reproduce any sort of crash using the original code. i refactored after I noticed i'd used localtime instead of gmtime by mistake and then the code was identical for win32 and non. 

I did read in the ms doc this snippet ```Both the 32-bit and 64-bit versions of gmtime, mktime, mkgmtime, and localtime all use one common tm structure per thread for the conversion``` could you have maybe had called this function on different threads causing a crash? This code is only called one time during log initialization so i'm inclined to leave it the same for both _WIN32 and not

inc_internal/zt_internal.h
```diff
@@ -246,6 +246,8 @@ int close_conn_internal(struct ziti_conn *conn);
 
 int establish_crypto(ziti_connection conn, message *msg);
 
+void ziti_log_format(char* time_str, size_t time_str_len, uv_timeval64_t start_time);
```
ekoby, line 249: I disagree with the name -- `ziti_fmt_time`?

library/ziti.c
```diff
@@ -179,15 +179,8 @@ int ziti_init_opts(ziti_options *options, uv_loop_t *loop, void *init_ctx) {
...
-    struct tm *start_tm = gmtime(&start_time.tv_sec);
     char time_str[32];
-#if _WIN32
-    SYSTEMTIME st;
-    GetLocalTime(&st);
-    sprintf(time_str, "%02d-%02d-%04dT%02d:%02d:%02d", st.wDay, st.wMonth, st.wYear, st.wHour, st.wMinute, st.wSecond);
-#else
-    strftime(time_str, sizeof(time_str), "%FT%T", start_tm);
-#endif
+    ziti_fmt_time(&time_str, sizeof(time_str), start_time);
```
scareything, line 183: "&" is technically not the correct way to pass an array as a pointer, because an array is a pointer. If you watch your build you'll probably see a warning on this line:

```
/Users/scarey/repos/github.com/openziti/ziti-sdk-c/library/ziti.c:183:19: warning: incompatible pointer types passing 'char (*)[32]' to parameter of type 'char *'
      [-Wincompatible-pointer-types]
    ziti_fmt_time(&time_str, sizeof(time_str), start_time);
                  ^~~~~~~~~
```

You either want `&time_str[0]` or just plain-old `time_str`.

PR #145, opened 2020-09-17
fix comparison of map fields

ekoby:


library/model_support.c
```diff
@@ -86,24 +88,23 @@ int model_cmp(void *lh, void *rh, type_meta *meta) {
...
+            rf_ptr = rf_addr;
 
-            if (ftm->comparer) {
-                rc = ftm->comparer(lf_ptr, rf_ptr);
-            }
-            else {
-                rc = model_cmp(lf_ptr, rf_ptr, ftm);
-            }
+            rc = model_map_compare(lf_ptr, rf_ptr, ftm);
         }
```
dovholuknf, line 106: might be nice to add a catchall 'else' at the end for the next time? not that there'll be a next time? :)

library/model_support.c
```diff
@@ -1009,6 +1010,43 @@ model_map_iter model_map_it_remove(model_map_iter it) {
     return next;
 }
 
+static int model_map_compare(model_map *lh, model_map *rh, type_meta *m) {
```
dovholuknf, line 1013: it feels like you really can't compare maps unless they're sorted, right? i don't think the map has a sort function does it?

dovholuknf (approved):

library/model_support.c
ekoby: sorting does not matter for comparison

dovholuknf: not for the function to execute properly - sure. but for it to be accurate to what a human would expect it does.

ekoby: a human would be wrong to make any assumptions wtr to map ordering

ekoby: also this is an internal function, not part of the API

PR #146, opened 2020-10-01
handle service.encryptionRequired

ekoby:


library/connect.c
```diff
@@ -552,10 +550,15 @@ void connect_reply_cb(void *ctx, message *msg) {
...
-                establish_crypto(conn, msg);
-                send_crypto_header(conn);
-                conn->state = Connected;
-                req->cb(conn, ZITI_OK);
+                int rc = ZITI_OK;
+                if (conn->encrypted) {
+                    rc = establish_crypto(conn, msg);
+                    if (rc == ZITI_OK) {
+                        send_crypto_header(conn);
+                    }
```
dovholuknf, line 558: should this return an error here?

dovholuknf: i see it is used in the cb below - maybe a log message then if the crypto establishment fails?

smilindave26 (approved):

PR #149, opened 2020-10-22
implement half-close support

### LINKED ISSUE #148: implement support for half close on ziti connections

ekoby:
[fixes #148]


smilindave26 (approved):

includes/ziti/ziti.h
```diff
@@ -518,6 +518,26 @@ extern int ziti_accept(ziti_connection clt, ziti_conn_cb cb, ziti_data_cb data_c
...
+ *
+ * @note this is roughly equivalent to calling `uv_shutdown()` on a duplex `uv_stream`, or
+ *       `shutdown(sock, SHUT_WR)` on socket fd.
+ *
+ * @param conn the #ziti_connection to be closed
+ *
+ * @return #ZITI_OK or corresponding #ZITI_ERRORS
+ */
+ZITI_FUNC
+extern int ziti_close_write(ziti_connection conn);
```
dovholuknf, line 539: i don't know about this, it seems odd to me. if i'm the one using this sdk why would i want to have `ziti_close_write`. Why wouldn't I just "not write to the connection" any more? is this function really needed?

library/ziti.c
```diff
@@ -381,6 +381,11 @@ int ziti_close(ziti_connection *conn) {
 }
 
 int ziti_write(ziti_connection conn, uint8_t *data, size_t length, ziti_write_cb write_cb, void *write_ctx) {
+    if (conn->state != Connected) {
+        ZITI_LOG(ERROR, "attempted write on conn[%d] in invalid state[%d]", conn->conn_id, conn->state);
```
dovholuknf, line 385: Shouldn't this check for `CloseWrite` too? I'd like to see a warning in the logs if i'm writing to a closed connection and maybe return a different value than 'invalid'?

includes/ziti/ziti.h
ekoby: the reason is to notify the other side that you're done

library/ziti.c
ekoby: not sure what you're getting at -- after you called ziti_close_write() you shouldn't be able to call ziti_write() on same connection

includes/ziti/ziti.h
ekoby: see matching PR for Go SDK https://github.com/openziti/sdk-golang/pull/99

library/ziti.c
dovholuknf: ugh. somehow my brain read `!= Connected` as 'disconnected' - i read it numerous times and still brain processed that as 'disconnected'... makes sense now though

library/connect.c
```diff
@@ -105,6 +107,11 @@ void on_write_completed(struct ziti_conn *conn, struct ziti_write_req_s *req, in
     }
 
     free(req);
+
+    if (conn->write_reqs == 0 && conn->state == CloseWrite) {
+        ZITI_LOG(DEBUG, "sending FIN");
```
dovholuknf, line 112: small nit but maybe this log statement belongs in the function so that if it's called elsewhere - it'll log?

dovholuknf (approved):

PR #150, opened 2020-10-27
adjust log level

ekoby:


programs/ziti-prox-c/proxy.c
```diff
@@ -181,7 +181,7 @@ static void on_ziti_write(ziti_connection conn, ssize_t status, void *ctx) {
 
 static void data_cb(uv_stream_t *stream, ssize_t nread, const uv_buf_t *buf) {
     struct client *clt = stream->data;
-    ZITI_LOG(INFO, "client[%s]: nread[%zd]", clt->addr_s, nread);
+    ZITI_LOG(TRACE, "client[%s]: nread[%zd]", clt->addr_s, nread);
```
dovholuknf, line 184: why not just remove it entirely? put it back if you need it? or - compile it in with an option?

smilindave26 (approved):

rentallect (approved):

programs/ziti-prox-c/proxy.c
dovholuknf: maybe we could have a ZITI_DEBUG function that's macro'ed out when not in dev?

PR #153, opened 2020-10-27
make sure we app gets data_cb with EOF after calling ziti_close_write()

ekoby:


library/connect.c
```diff
@@ -60,7 +60,6 @@ static void free_conn_req(struct ziti_conn_req *r) {
 };
 
 int close_conn_internal(struct ziti_conn *conn) {
-    ZITI_LOG(TRACE, "connection[%d]", conn->conn_id);
```
dovholuknf, line 63: YES!!!! 👍

dovholuknf (approved):

Issue #154, opened 2020-10-28
rework ziti_close() and ziti_connection lifecycle


ekoby:
fix/improve constency of ziti_connection lifecycle:
- currently ZITI_EOF must be handled differently from other errors by the application code
- may need to require app to call ziti_close() even after receiving error on ziti_conn
- ziti_close() with a callback -- make it closer to libuv standards

smilindave26:
+1 on ziti_close() with a callback.  There are some gyrations in some of the Apple stuff to work around this...

PR #155, opened 2020-10-28
dont use EOF when connection is terminated

ekoby:
this a quick fix -- better solution is outlined in #154 

for now: since ZITI_EOF indicates that connection is still alive (per half-close feature) don't use it if connection state is `Closed`

### DIFF
--- a/tests/test_ziti_model.cpp
+++ b/tests/test_ziti_model.cpp
@@ -337,6 +337,67 @@ TEST_CASE("service config test", "[model]") {
     free_ziti_intercept(&cfg);
 }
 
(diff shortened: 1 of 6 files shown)


tests/test_ziti_model.cpp
```diff
@@ -337,6 +337,67 @@ TEST_CASE("service config test", "[model]") {
     free_ziti_intercept(&cfg);
 }
 
+TEST_CASE("config change", "[model]") {
```
smilindave26, line 340: :)

smilindave26 (approved):

PR #156, opened 2020-10-29
Implement router selection

ekoby:
use latency to select best channel for connection

also fixes:
prevent writing to closed client(tcp) handle
prevent ziti_disconnect_async firing on disposed connection

### DIFF
--- a/inc_internal/edge_protocol.h
+++ b/inc_internal/edge_protocol.h
@@ -54,6 +54,7 @@ enum header_id {
     ReflectedHeaderBitMask = 1 << 7,
     MaxReflectedHeader = (1 << 8) - 1,
 
+    LatencyProbeTime = 128,
     UUIDHeader = 128,
 
     ConnIdHeader = 1000,
--- a/library/channel.c
+++ b/library/channel.c
@@ -113,11 +113,7 @@ int ziti_channel_close(ziti_channel_t *ch) {
     return r;
 }
 
-int ziti_channel_connect(ziti_context ztx, const char *name, const char *url, ch_connect_cb cb, void *cb_ctx) {
-    size_t ch_name_len = strlen(name) + strlen(url) + 2;
-    char *ch_name = malloc(ch_name_len);
-    snprintf(ch_name, ch_name_len, "%s@%s", name, url);
-
+int ziti_channel_connect(ziti_context ztx, const char *ch_name, const char *url, ch_connect_cb cb, void *cb_ctx) {
     ziti_channel_t *ch = model_map_get(&ztx->channels, ch_name);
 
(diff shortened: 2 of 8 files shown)


library/channel.c
```diff
@@ -134,14 +138,14 @@ int ziti_channel_connect(ziti_context ztx, const char *url, ch_connect_cb cb, vo
             ZITI_LOG(ERROR, "should not be here: %s", ziti_errorstr(ZITI_WTF));
             return ZITI_WTF;
```
dovholuknf, line 139: ch_name was malloc'ed but not free'ed here. probably no big deal considering that it's in this else...

library/connect.c
```diff
@@ -191,10 +194,31 @@ static int ziti_connect(struct ziti_ctx *ctx, const ziti_net_session *session, s
     conn->token = session->token;
 
     ziti_edge_router **er;
+    ziti_channel_t *best_ch = NULL;
+    uint64_t best_latency = UINT64_MAX;
+
+    char ch_name[128];
     for (er = session->edge_routers; *er != NULL; er++) {
+        snprintf(ch_name, sizeof(ch_name), "%s@%s", (*er)->name, (*er)->ingress.tls);
```
dovholuknf, line 202: is 128 bytes enough guaranteed to always be enough? can't calculate the size?

library/connect.c
```diff
@@ -191,10 +194,31 @@ static int ziti_connect(struct ziti_ctx *ctx, const ziti_net_session *session, s
...
+        }
+    }
+
+    if (best_ch) {
+        ZITI_LOG(DEBUG, "selected ch[%s] for best latency(%ldms)", best_ch->ingress, best_ch->latency);
         req->chan_tries++;
-        ZITI_LOG(TRACE, "connecting to %s(%s) for session[%s]", (*er)->name, (*er)->ingress.tls, conn->token);
-        ziti_channel_connect(ctx, (*er)->ingress.tls, on_channel_connected, req);
+        on_channel_connected(best_ch, req, ZITI_OK);
     }
```
dovholuknf, line 222: seems like a warn if there's no best channel selected? why keep the req->chan_tries line?

dovholuknf (approved):

inc_internal/edge_protocol.h
```diff
@@ -54,6 +54,7 @@ enum header_id {
     ReflectedHeaderBitMask = 1 << 7,
     MaxReflectedHeader = (1 << 8) - 1,
 
+    LatencyProbeTime = 128,
```
smilindave26, line 57: worth a comment on why it's ok to share 128?

smilindave26 (approved):

PR #162, opened 2020-11-07
detect null arrays

andrewpmartinez:


ekoby:
Looks good, please add a test for it

ekoby (approved):
Just needs some tests for it

library/model_support.c
```diff
@@ -381,6 +381,11 @@ void model_free(void *obj, type_meta *meta) {
 }
 
 static int parse_array(void **arr, const char *json, jsmntok_t *tok, type_meta *el_meta) {
+    if(tok-> type == JSMN_PRIMITIVE && json[tok->start] == 'n'){ //null check
```
scareything, line 384: Is it sufficient to look only at the first character of the token here? My instinct is to use strncmp(&json[tok->start], "null", 4)

scareything (approved):

andrewpmartinez:
> Just needs some tests for it

Added tests. Since I went through the trouble of learning how I added the service model for posture checks.

library/model_support.c
andrewpmartinez: Yes because jsmn defines JSMN_STRING for strings and JSMN_PRIMITIVE for bool, null, and numbers. It is honestly an interesting approach: https://github.com/zserge/jsmn#api

```
Note: Unlike JSON data types, primitive tokens are not divided into numbers, booleans and null, because one can easily tell the type using the first character:

't', 'f' - boolean
'n' - null
'-', '0'..'9' - number
```

PR #174, opened 2020-11-18
lookup and log edge router version

ekoby:


library/channel.c
```diff
@@ -482,7 +482,11 @@ static void hello_reply_cb(void *ctx, message *msg) {
 
     int cb_code = ZITI_OK;
     if (success) {
-        ZITI_LOG(TRACE, "channel[%d] connected: %d", ch->id, success);
+        uint8_t *erVersion = "<unknown>";
```
smilindave26, line 485: this NULL terminates correctly/reliably (so the strlen on next line is ok)?

ekoby: yes, strlen() returns the length without `\0` byte

smilindave26: yeah - just checking "a quoted string" includes the trailing '\0'.  I looked it up to make sure: "A byte or code of value zero is appended to each multibyte character sequence that results from a string literal or literals."  :)

smilindave26 (approved):

PR #175, opened 2020-11-19
set source_identity on incoming connections

scareything:
Make use of the CallerIdHeader when the client sends it.


library/connect.c
```diff
@@ -818,7 +820,7 @@ int ziti_channel_start_connection(struct ziti_conn *conn) {
             },
             {
                     .header_id = CallerIdHeader,
-                    .length = strlen(conn->ziti_ctx->session->identity->name),
+                    .length = strlen(conn->ziti_ctx->session->identity->name) + 1,
```
ekoby, line 823: this is not needed -- we shouldn't put '\0' byte into the header, it does not translate into other languages correctly

library/connect.c
```diff
@@ -1080,6 +1082,12 @@ static void process_edge_message(struct ziti_conn *conn, message *msg, int code)
             message_get_bytes_header(msg, AppDataHeader, &app_data, &app_data_sz);
             ziti_connection clt;
             ziti_conn_init(conn->ziti_ctx, &clt, app_data);
+            uint8_t *source_identity = NULL;
+            size_t source_identity_sz = 0;
+            bool caller_id_sent = message_get_bytes_header(msg, CallerIdHeader, &source_identity, &source_identity_sz);
```
ekoby, line 1087: should allocate extra byte here instead (see above comment)

ekoby (changes requested):
we should not add and extra byte on string headers (identity, caller id) -- it is inconsistent with other implementations (Go, Java)

ekoby (approved):

PR #178, opened 2020-11-23
Fast model map

ekoby:
improve `model_map` performance for set/get operations

### DIFF
--- a/library/model_support.c
+++ b/library/model_support.c
@@ -907,7 +907,7 @@ static void _free_tag(tag *t) {
 
 struct model_map_entry {
     char *key;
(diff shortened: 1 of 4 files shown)


library/model_support.c
```diff
@@ -905,76 +905,135 @@ static void _free_tag(tag *t) {
     }
 }
 
+//#if defined()
```
scareything, line 908: cruft?

scareything (approved):

library/model_support.c
ekoby: removed

PR #182, opened 2020-11-24
fix ziti-tunneler build

dovholuknf:


scareything (approved):

includes/ziti/ziti_log.h
```diff
@@ -5,6 +5,8 @@
 #ifndef ZITI_SDK_ZITI_LOG_H
 #define ZITI_SDK_ZITI_LOG_H
 
+#include <uv.h>
```
dovholuknf, line 8:
```
ZITI_FUNC extern void init_debug(uv_loop_t *loop);
```
brought uv to the party

ekoby (approved):

PR #191, opened 2020-11-27
implement channel re-connect

ekoby:
channels are persisted on disconnect, or failure to connect. 
each failed channel will attempt to re-connect with exponential backoff

### DIFF
--- a/library/channel.c
+++ b/library/channel.c
@@ -575,7 +575,11 @@ static void reconnect_channel(ziti_channel_t *ch) {
     uv_timer_init(ch->ctx->loop, t);
     t->data = ch;
 
-    unsigned int backoff = rand() % MIN(ch->reconnect_count, MAX_BACKOFF);
+    int count = ch->reconnect_count;
+    if (count > MAX_BACKOFF) {
+        count = MAX_BACKOFF;
+    }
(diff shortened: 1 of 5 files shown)


library/channel.c
```diff
@@ -532,16 +553,57 @@ static void async_write(uv_async_t* ar) {
...
+static void reconnect_channel(ziti_channel_t *ch) {
+    ch->reconnect_count++;
+    uv_timer_t *t = malloc(sizeof(uv_timer_t));
+    uv_timer_init(ch->ctx->loop, t);
+    t->data = ch;
+
+    unsigned int backoff = rand() % MIN(ch->reconnect_count, MAX_BACKOFF);
+
+    uint64_t timeout = (1U << backoff) * BACKOFF_TIME;
+    ZITI_LOG(INFO, "ch[%d] reconnecting in %ld ms (attempt = %d)", ch->id, timeout, ch->reconnect_count);
```
dovholuknf, line 581: seems like debug more than info to me?

dovholuknf (approved):

library/channel.c
ekoby: They should be fairly rare

PR #192, opened 2020-11-30
logging API

ekoby:


library/utils.c
```diff
@@ -126,15 +122,45 @@ static char elapsed_buffer[32];
...
+        ziti_log_set_logger(logger ? logger : default_log_writer);
+    } else {
+        ziti_log_set_logger(log_func);
+    }
 }
 
-void init_debug(uv_loop_t *loop) {
+void ziti_log_set_level(int level) {
+    if (level == ZITI_LOG_DEFAULT_LEVEL) {
+        char *lvl = getenv("ZITI_LOG");
```
dovholuknf, line 146: i feel like this would be hard to track down if didn't know about it. maybe this should produce a log message of its own as to why the log level is set. or - why not just drop ZITI_LOG entirely?

includes/ziti/ziti_log.h
```diff
@@ -40,23 +40,32 @@ enum DebugLevel {
...
 typedef void (*log_writer)(int level, const char *loc, const char *msg, size_t msglen);
 
 ZITI_FUNC extern void
 ziti_logger(int level, const char *file, unsigned int line, const char *func, const char *fmt, ...);
 
-ZITI_FUNC extern void init_debug(uv_loop_t *loop);
+// call once
+// use ZITI_LOG_DEFAULT_LEVEL to use default(INFO)/ZITI_LOG env var
+// pass logger = NULL to use default output
+ZITI_FUNC extern void ziti_log_init(uv_loop_t *loop, int level, log_writer logger);
```
dovholuknf, line 60: probably needs comments as to when this needs to be called. do i need to ever call this? If not i assume it's just no-ops all the way down as "default output" or is it stderr/out for 'default' output? is it smart for me to call this before ziti_init? (i would think)

dovholuknf (approved):

library/utils.c
```diff
@@ -172,10 +194,10 @@ void ziti_logger(int level, const char *file, unsigned int line, const char *fun
     va_start(argp, fmt);
     char location[128];
     if (func && func[0]) {
-        snprintf(location, sizeof(location), "%s:%d %s()", file, line, func);
+        snprintf(location, sizeof(location), "%s:%d %s()", file + SOURCE_PATH_SIZE, line, func);
```
smilindave26, line 197: I liked having the `ZITI_LOG_PREFIX` in the message (so you'd see stuff like `[2020-12-01T15:19:32.744Z] INFO ziti-tunnel-sdk-c:../lib/ziti_tunnel.c:44 ziti_tunneler_init(): Ziti Tunneler SDK (v0.7.17)`)

smilindave26 (approved):

includes/ziti/ziti_log.h
```diff
@@ -40,23 +40,32 @@ enum DebugLevel {
...
 
+#define ZITI_LOG_DEFAULT_LEVEL (-1)
+
 typedef void (*log_writer)(int level, const char *loc, const char *msg, size_t msglen);
 
 ZITI_FUNC extern void
 ziti_logger(int level, const char *file, unsigned int line, const char *func, const char *fmt, ...);
 
-ZITI_FUNC extern void init_debug(uv_loop_t *loop);
+// call once
```
smilindave26, line 57: will be called for each identity in the ZDE/ZME apps

library/utils.c
ekoby: it seemed redundant to me, we get full relative path of the files now so it is more clear where the file is coming from

includes/ziti/ziti_log.h
ekoby: that is fine, important init is only done once

smilindave26: `ziti_log_init` will be called for each identity (in `ziti_init_ops()`)

PR #202, opened 2020-12-10
Fix server 2016 crash

dovholuknf:
This change will export the __USE_MINGW_ANSI_STDIO macro for MINGW users. This will prevent older msvcrt.dll users from having crashes when logging size_t/ssize_t along with strings


ekoby (approved):

ekoby:
I think it would be better to set it at `CMAKE_C_FLAGS` level

PR #206, opened 2020-12-17
Ziti conn improvement

ekoby:


includes/ziti/ziti.h
```diff
@@ -689,12 +698,16 @@ extern int ziti_accept(ziti_connection clt, ziti_conn_cb cb, ziti_data_cb data_c
  * When no longer needed a [connection](#ziti_connection) should be closed to gracefully disconnect. This
  * function should be invoked after any status is returned which indicates an error situation.
  *
+ * This method ensures initiates disconnect(if needed) and release all associated resources.
+ * After close_cb() is called ziti_connection handle is no longer valid.
```
dovholuknf, line 702: After close_cb() is called, **the** ziti_connection handle is no longer valid.

includes/ziti/ziti.h
```diff
@@ -689,12 +698,16 @@ extern int ziti_accept(ziti_connection clt, ziti_conn_cb cb, ziti_data_cb data_c
  * When no longer needed a [connection](#ziti_connection) should be closed to gracefully disconnect. This
  * function should be invoked after any status is returned which indicates an error situation.
  *
+ * This method ensures initiates disconnect(if needed) and release all associated resources.
```
dovholuknf, line 701: > This method ensures initiates disconnect
needs some new words at the least but really isn't it sort of stated in the lines above? i started to rework it but - then realized i don't really have the time to offer anything insightful just yet.... :(

library/channel.c
```diff
@@ -30,6 +30,14 @@ limitations under the License.
 #define BACKOFF_TIME 3000 /* 3 seconds */
 #define MAX_BACKOFF 5 /* max reconnection timeout: (1 << 5) * BACKOFF_TIME = 96 seconds */
 
+enum ChannelState {
+    Initial,
```
dovholuknf, line 34: Initial -> Initialized to stay consistent with "Connected" and "Disconnected"?

dovholuknf: is there a "1/2 closed" state to add in?

dovholuknf: and why wouldn't this be in ziti.h? just not exposed yet?

library/connect.c
```diff
@@ -22,13 +22,35 @@ limitations under the License.
...
 #include "win32_compat.h"
 
-static const char *TYPE_BIND = "Bind";
-static const char *TYPE_DIAL = "Dial";
 static const char *INVALID_SESSION = "Invalid Session";
 static const int MAX_CONNECT_RETRY = 3;
 
 #define crypto(func) crypto_secretstream_xchacha20poly1305_##func
 
+#define conn_states(XX) \
```
dovholuknf, line 30: these states don't overlap entirely ChannelState. i expect that is that on purpose but it's not clear to me as to why that would be. as long as that's on purpose and i just don't understand 'why' that's fine... if you can add clarity around that in the form of code comments that might be a good idea?

library/connect.c
```diff
@@ -50,12 +73,21 @@ static int send_fin_message(ziti_connection conn);
...
 static void free_handle(uv_handle_t *h) {
     free(h);
 }
 
+const char* ziti_conn_state(ziti_connection conn) {
+    return conn ? conn_state_str[conn->state] : "<NULL>";
+}
+
+static void conn_set_state(struct ziti_conn *conn, enum conn_state state) {
+    ZITI_LOG(VERBOSE, "conn[%d] transitioning %s => %s", conn->conn_id, conn_state_str[conn->state], conn_state_str[state]);
```
dovholuknf, line 87: why not use `ziti_conn_state` here and avoid any null pointer issue possibility?

library/connect.c
ekoby: it is on purpose, some conn states make no sense for channels: Bound(ing), Accepting, HalfClose, etc.

states are now private to corresponding files, so that should eliminate confusion

library/ziti.c
```diff
@@ -390,7 +369,7 @@ int ziti_conn_init(ziti_context ztx, ziti_connection *conn, void *data) {
     c->ziti_ctx = ztx;
     c->data = data;
     c->channel = NULL;
-    c->state = Initial;
+//    c->state = Initial;
```
dovholuknf, line 372: when i see a stray comment like this i wonder if it was intentional. drawing your eyes here just to check

dovholuknf (approved):
approved pending minor doc updates

PR #212, opened 2021-01-12
extend ziti_client_cb API for better receiving of peer supplied data

ekoby:


programs/sample-host/sample-host.c
```diff
@@ -58,17 +58,17 @@ static void on_client_connect(ziti_connection clt, int status) {
     }
 }
 
-static void on_client(ziti_connection serv, ziti_connection client, int status) {
+static void on_client(ziti_connection serv, ziti_connection client, int status, ziti_client_ctx *clt_ctx) {
     if (status == ZITI_OK) {
         const char *source_identity = ziti_conn_source_identity(client);
```
scareything, line 63: should this sample show using ->caller_id?

includes/ziti/ziti.h
```diff
@@ -330,10 +336,11 @@ typedef void (*ziti_conn_cb)(ziti_connection conn, int status);
  * @param serv hosting connection, initialized with ziti_listen()
  * @param client client connection - generally passed to ziti_accept() in this function
  * @param status #ZITI_OK or error
+ * @param ctx object containing application data passed by dialing identity, \see ziti_dial_opts
```
scareything, line 339: it may be obvious, but should this mention that the contents of `ctx` are only valid in the client_cb?

scareything (approved):

programs/sample-host/sample-host.c
ekoby: done

includes/ziti/ziti.h
ekoby: done

PR #213, opened 2021-01-14
check if update needed before service refresh

ekoby:


library/ziti.c
```diff
@@ -610,9 +610,29 @@ static void update_services(ziti_service_array services, ziti_error *error, void
     model_map_clear(&updates, NULL);
 }
 
+static void check_service_update(ziti_service_update *update, ziti_error *err, void *ctx) {
+    ziti_context ztx = ctx;
+    bool need_update = false;
+
+    if (err) { // API not supported - do refresh
+        need_update = true;
```
andrewpmartinez, line 618: If this does succeed you don't have to hit any other API endpoint to keep an API Session alive.'

API Sessions are marked alive on any authenticated request or on any authenticated connection to an Edge Router.

I'd move the check to see what mode the SDK operates in to right after authentication and then set a flag to do the right behavior.

ekoby: I am not sure what you mean, the service refresh is scheduled after successful auth

andrewpmartinez: If this is working with an old controller it will always execute two requests to update services (1 to check if the new service-updates endpoint exists and 1 to get the services) on every polling interval. Persisting the result of whether the `service-updates` endpoint allows us to issue only 1 request for an older controller.

Also, you mentioned yesterday that the controller was still trying to keep the API Session alive via `GET /current-api-session`. However, service polling does that as well. So requests to `GET /current-api-session` are not needed.

andrewpmartinez (approved):

PR #215, opened 2021-01-19
add app info and submit it to controller

ekoby:


library/config.c
```diff
@@ -24,6 +24,14 @@ limitations under the License.
 
 const char* ZITI_SDK_CONFIG = "ZITI_SDK_CONFIG";
 
+const char* APP_ID = NULL;
+const char* APP_VERSION = NULL;
+
+void ziti_set_app_info(const char *app_id, const char *app_version) {
+    APP_ID = strdup(app_id);
```
dovholuknf, line 31: i guess i shouldn't call this function 1000's of times right? :)

dovholuknf (approved):
possible memory leak i think if someone called ziti_set_app_info a lot, right?

library/config.c
ekoby: updated, knock yourself out

PR #218, opened 2021-01-22
make ziti_dump more flexible

ekoby:


includes/ziti/ziti.h
```diff
@@ -555,10 +555,17 @@ int ziti_ctx_free(ziti_context *ctxp) ;
  * This function will output debugging information to standard out. The output from this command may
  * be useful when submitting issues.
  *
+ * this method is designed to be suitable to use with `fprintf()` like this:
```
dovholuknf, line 558: out of curiosity - why not use the logger callback if it's declared?

ekoby: it is for the app to use, not necessarily want to send this to log

dovholuknf (approved):

PR #222, opened 2021-01-26
skip channel creation to 'ws' edge router bindings

rentallect:


library/connect.c
```diff
@@ -323,22 +323,29 @@ static int ziti_connect(struct ziti_ctx *ctx, const ziti_net_session *session, s
...
+                }
             }
+            else {
+                ZITI_LOG(TRACE, "connecting to %s(%s) for session[%s]", (*er)->name, (*er)->ingress.tls, conn->token);
+                ziti_channel_connect(ctx, ch_name, (*er)->ingress.tls, on_channel_connected, conn);
+            }
+            free(ch_name);
+        } else if ((*er)->ingress.ws) {
+            size_t ch_name_len = strlen((*er)->name) + strlen((*er)->ingress.ws) + 2;
+            char *ch_name = malloc(ch_name_len);
```
scareything, line 345: This buffer could be avoided by passing `(*er)->name, (*er)->ingress.ws` directly into the ZITI_LOG below.
If you do want to use a separate buffer though, the buffer needs to be free'd after being used in the ZITI_LOG.

PR #226, opened 2021-01-29
reworks posture data submission

andrewpmartinez:
- sent via bulk update, 1 http POST instead of 1 per response
- only send data on change

No external SDK API Changes.

### DIFF
--- a/library/posture.c
+++ b/library/posture.c
@@ -49,7 +49,7 @@ extern void posture_init(struct ziti_ctx *ztx, long interval_secs) {
         pc->previous_session_id = NULL;
         pc->must_send_every_time = true;
         pc->must_send = false;
-        pc->bulk_supported = false;
+        pc->bulk_supported = true;
         ztx->posture_checks = pc;
     }
 
@@ -285,7 +285,6 @@ static void ziti_pr_post_bulk_cb(void *obj, ziti_error *err, void *ctx) {
         ztx->posture_checks->must_send = true; //error, must try again
         if (err->http_code == 404) {
             ztx->posture_checks->bulk_supported = false;
-
         }
     } else {
         ztx->posture_checks->must_send = false; //did not error, can skip submissions
--- a/library/posture.c
+++ b/library/posture.c
@@ -27,17 +27,42 @@ static void ziti_handle_mac(ziti_context ztx, char *id, char **mac_addresses, in
 
 static void ziti_handle_domain(ziti_context ztx, char *id, char *domain);
 
-static void ziti_handle_os(ziti_context ztx, char *id, char *os_type, char *os_version, char *os_build);
+static void ziti_handle_os(ziti_context ztx, const char *id, char *os_type, char *os_version, char *os_build);
+
+static void ziti_send_pr(ziti_context ztx);
+
+static void ziti_send_pr_bulk(ziti_context ztx);
+
+static void ziti_send_pr_individually(ziti_context ztx);
 
 static void
-ziti_handle_process(ziti_context ztx, char *id, char *path, bool is_running, char *sha_512_hash, char **signers,
+ziti_handle_process(ziti_context ztx, const char *id, char *path, bool is_running, char *sha_512_hash, char **signers,
                     int num_signers);
 
 extern void posture_init(struct ziti_ctx *ztx, long interval_secs) {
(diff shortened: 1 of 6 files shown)


andrewpmartinez:
I will revert formatting changes. Apparently, my habit of constantly formatting go files gets me into trouble here.

ekoby:
You need to handle absence of bulk submit API and revert to individual submissions

library/posture.c
```diff
@@ -165,29 +194,164 @@ void ziti_send_posture_data(struct ziti_ctx *ztx) {
...
+            }
+        }
+    } //else is_sent was defaulted to true
+
+    model_map_set(ztx->posture_checks->current_responses, pr_obj_key, current_info);
+}
+
+static void ziti_pr_post_bulk_cb(void *obj, ziti_error *err, void *ctx) {
+    if (err != NULL) {
+        ZITI_LOG(ERROR, "error during bulk posture response submission (%d) %s", err->http_code, err->message);
```
ekoby, line 226: fallback to individual submissions?

ekoby (changes requested):

library/posture.c
andrewpmartinez: Added just a few lines below

library/posture.c
```diff
@@ -21,22 +21,35 @@ limitations under the License.
 #define NANOS(s) ((s) * 1e9)
 #define MILLIS(s) ((s) * 1000)
 
+const int NO_TIMEOUTS = -1;
```
andrewpmartinez, line 24: I added a sentinel value in the controller for configurable timeouts. If it is the `timeout` property is set to -1 then:

1. time variability is supported
2. there are no timeouts on the posture checks if the timeout == -1
2)

andrewpmartinez: So if 

new controller + old SDK, the old SDK will over submit posture data, but there should be no issue with that

old controller + new SDK, the new SDK will detect the old controller not supporting timeouts and will submit as the old SDKs used to.

library/posture.c
```diff
@@ -165,29 +237,219 @@ void ziti_send_posture_data(struct ziti_ctx *ztx) {
...
+    model_map_set(ztx->posture_checks->current_responses, pr_obj_key, current_info);
+}
+
+static void ziti_pr_post_bulk_cb(void *obj, ziti_error *err, void *ctx) {
+    ziti_context ztx = ctx;
+    if (err != NULL) {
+        ZITI_LOG(ERROR, "error during bulk posture response submission (%d) %s", err->http_code, err->message);
+        ztx->posture_checks->must_send = true; //error, must try again
+        if (err->http_code == 404) {
+            ztx->posture_checks->bulk_supported = false;
```
ekoby, line 287: I would like to consolidate controller features into one place to be able to reset them when controller version changes (and not requiring re-init/restart)

ziti_context already has one flag here https://github.com/openziti/ziti-sdk-c/blob/main/inc_internal/zt_internal.h#L181

ekoby (approved):

library/posture.c
```diff
@@ -21,22 +21,35 @@ limitations under the License.
...
 
 extern void posture_init(struct ziti_ctx *ztx, long interval_secs) {
     if (ztx->posture_checks == NULL) {
         NEWP(pc, struct posture_checks);
+
         pc->interval = (double) interval_secs;
+        pc->previous_session_id = NULL;
+        pc->must_send_every_time = true;
+        pc->must_send = false;
+        pc->bulk_supported = false;
```
ekoby, line 52: should it be set to `true`?

it is set to false on 404, and I don't see when it becomes `true`

andrewpmartinez: That was me testing the individual posting support. It should be `true`

library/posture.c
andrewpmartinez: Is the only thing you want for me to move my `bool` there ?

library/posture.c
andrewpmartinez: Fixed

ekoby (approved):

library/posture.c
andrewpmartinez: Move and swapped it to `no_bulk_posture_response_api` to match `no_service_updates_api`

PR #227, opened 2021-01-29
improve ziti context and connection logging

ekoby:
Ziti context relevant logs will have prefix `ztx[id]` 
ziti connection logging will have prefix `conn[{ztx.id}.{conn.id}]`

### DIFF
--- a/library/ziti.c
+++ b/library/ziti.c
@@ -41,6 +41,8 @@ limitations under the License.
 #endif
 #endif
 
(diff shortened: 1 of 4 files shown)


library/ziti.c
```diff
@@ -318,6 +323,9 @@ int ziti_shutdown(ziti_context ztx) {
 }
 
 int ziti_ctx_free(ziti_context *ctxp) {
+    if (ctxp == NULL || *ctxp == NULL) return ZITI_OK;
```
dovholuknf, line 326: is this something that would be worth logging and tracking down?

dovholuknf (approved):

PR #230, opened 2021-02-02
Avoid using expired session

ekoby:


library/ziti.c
```diff
@@ -712,12 +726,20 @@ static void session_cb(ziti_session *session, ziti_error *err, void *ctx) {
         free_ziti_session(old_session);
         FREE(old_session);
 
+        uv_timeval64_t now;
+        uv_gettimeofday(&now);
+
+        int time_diff = (int) (now.tv_sec - session->updated->tv_sec);
+        if (abs(time_diff) > 10) {
```
dovholuknf, line 733: why not > 0? why the leeway?

library/ziti.c
```diff
@@ -712,12 +726,20 @@ static void session_cb(ziti_session *session, ziti_error *err, void *ctx) {
...
+        }
 
         if (session->expires) {
-            uv_timeval64_t now;
-            uv_gettimeofday(&now);
+            // adjust expiration to local time if needed
+            session->expires->tv_sec += time_diff;
             ZTX_LOG(DEBUG, "ziti API session expires in %ld seconds", (long) (session->expires->tv_sec - now.tv_sec));
-            long delay = (session->expires->tv_sec - now.tv_sec) * 3 / 4;
+            long delay = (session->expires->tv_sec - now.tv_sec) - 10;
```
dovholuknf, line 742: is this on purpose? the - 10?  10 just seems arbitrary ?

ekoby: consistent with golang SDK

library/ziti.c
ekoby: it is just an informational error message, we have no control of how often host time is synced with UTC, giving it leeway avoids logging this too often

smilindave26 (approved):

PR #240, opened 2021-02-25
inital mfa

andrewpmartinez:


library/auth_queries.c
```diff
@@ -0,0 +1,323 @@
...
+
+void ziti_auth_query_mfa_auth_internal_cb(void *empty, ziti_error *err, void *ctx) {
+    ziti_mfa_auth_ctx *mfa_ctx = ctx;
+    ziti_context ztx = mfa_ctx->ztx;
+    if (err != NULL) {
+        ZITI_LOG(ERROR, "error during MFA auth call: %d - %s - %s", err->http_code, err->code, err->message);
+        mfa_ctx->attempts++;
+
+        if (mfa_ctx->attempts >= MFA_MAX_ATTEMPTS) {
+            exit(err->err);
```
andrewpmartinez, line 260: Wasn't sure what to do if someone just fails over and over to put the right TOTP/recovery code in.

scareything: alt text

ekoby: this is bad idea when we allow multiple identities

andrewpmartinez: I simply removed it. The external application can track attempts and let it decide how it wants to handle the situation (including giving up on the user).

programs/ziti-prox-test-c/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+    target_include_directories(ziti-prox-test-c PRIVATE win32/include)
+    target_sources(ziti-prox-test-c PRIVATE win32/src/getopt.c)
+endif()
+
+target_link_libraries(ziti-prox-test-c PUBLIC ziti subcommand)
+target_include_directories(ziti-prox-test-c PRIVATE ${ziti-sdk_SOURCE_DIR}/inc_internal)
+
+set(ZITI_ARCHIVE_NAME "ziti-prox-test-c-${ZITI_VERSION}-${CPACK_SYSTEM_NAME}.${archive_sfx}")
+set(ZITI_EXECUTABLE_NAME "ziti-prox-test-c${CMAKE_EXECUTABLE_SUFFIX}")
+if (UNIX)
```
ekoby, line 15: do we really need this packed?

andrewpmartinez: Removed.

includes/ziti/ziti_model.h
```diff
@@ -20,6 +20,15 @@ limitations under the License.
 
 #include "model_support.h"
 
+#define ZITI_AUTH_QUERY_MFA_MODEL(XX, ...) \
+XX(type_id, string, none, typeId, __VA_ARGS__) \
```
ekoby, line 24: is this an enum? also `provider`?

andrewpmartinez: It will be, right now we have only 1 value.

dovholuknf (approved):

PR #244, opened 2021-03-03
parse \\u JSON escaped characters, more valid JSON output

ekoby:


tests/model_tests.cpp
```diff
@@ -488,3 +488,22 @@ TEST_CASE("map compare", "[model]") {
     free_Basket(&b1);
     free_Basket(&b2);
 }
+
+TEST_CASE("parse-json-u-escape", "[model]") {
```
dovholuknf, line 492: maybe a test case for when the json escaped string is malformed?

dovholuknf (approved):

PR #250, opened 2021-03-10
fix multiple posture submits erroring on top of eachother

andrewpmartinez:
- fix memory leaks on model maps and ziti_error's in posture.c
- avoid using shared memory during individual posture response submits
  for older controllers

### DIFF
--- a/library/posture.c
+++ b/library/posture.c
@@ -82,7 +82,7 @@ static void ziti_pr_free_pr_cb_ctx(pr_cb_ctx *ctx) {
     FREE(ctx)
 }
 
-extern void ziti_posture_init(ziti_context ztx, long interval_secs) {
+void ziti_posture_init(ziti_context ztx, long interval_secs) {
     if (ztx->posture_checks == NULL) {
         NEWP(pc, struct posture_checks);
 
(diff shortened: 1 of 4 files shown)


library/posture.c
```diff
@@ -15,80 +15,123 @@ limitations under the License.
...
         uv_unref((uv_handle_t *) &ztx->posture_checks->timer);
     }
 }
 
 extern void ziti_posture_checks_free(struct posture_checks *pcs) {
-    model_map_clear(pcs->previous_responses, NULL);
-    model_map_clear(pcs->current_responses, NULL);
-    FREE(pcs);
-}
+    if (pcs != NULL) {
```
ekoby, line 109: it is not safe if pcs.timer is not closed

andrewpmartinez: Closed? Stopped? 

I was following the pattern I saw in `ziti.c` for `ziti_shutdown()` -> `ziti_ctx_free()`.

Where the `ziti_shutdown` I assumed would always be called first followed by a `ziti_ctx_free`.

andrewpmartinez: I can blindly call `uv_timer_stop` here though. It checks the state first to see if it needs stopping.

andrewpmartinez: updated

ekoby: stop is not enough, I believe, it's still referenced by the loop. you need to call `uv_close()`. freeing memory is not safe until close_cb is called (you can do it inside close_cb)

andrewpmartinez: I added `ziti_posture_checks_shutdown()` and call it before anything is freed (one line below this comment string).

andrewpmartinez: Fixed?

ekoby (approved):
LGTM

PR #254, opened 2021-03-12
on dupe info, don't use old pointers

andrewpmartinez:


library/posture.c
```diff
@@ -442,6 +442,9 @@ static void ziti_pr_send_individually(ziti_context ztx) {
             NEWP(new_info, pr_info);
             memcpy(new_info, info, sizeof(pr_info));
```
andrewpmartinez, line 443: memcopy was copying pointers too =-(

dovholuknf (approved):

PR #255, opened 2021-03-12
better model to json

ekoby:
replaces awkward to use `json_from_XXXX()` with `char* XXX_to_json()`. Freeing of returned string is responsibility of the caller
also adds `ssize_t XXX_to_json_r()` for fixed buffer output returns -1 if output buffer is too small for given model object

### DIFF
--- a/includes/ziti/model_support.h
+++ b/includes/ziti/model_support.h
@@ -52,6 +52,8 @@ limitations under the License.
 
 #define MODEL_API
 
+#define MODEL_JSON_COMPACT 0x1
+
 #define none(t) t
 #define ptr(t)  t*
 #define array(t) t##_array
--- a/includes/ziti/model_support.h
+++ b/includes/ziti/model_support.h
@@ -28,6 +28,16 @@ limitations under the License.
 
 #include "externs.h"
 
+#if !defined(__DEFINED_ssize_t) && !defined(__ssize_t_defined)
+#if _WIN32
+typedef intptr_t ssize_t;
+#define __DEFINED_ssize_t
+#define __ssize_t_defined
+#else
+#include <unistd.h>
+#endif
+#endif
+
 /**
  * set of macros to help generate struct and function for our model;
  *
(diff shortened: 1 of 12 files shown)


includes/ziti/model_support.h
```diff
@@ -71,8 +71,8 @@ MODEL_API int cmp_##type(const type *lh, const type *rh); \
 MODEL_API void free_##type##_array(array(type) *ap);\
 MODEL_API int parse_##type(ptr(type) v, const char* json, size_t len);\
 MODEL_API int parse_##type##_ptr(ptr(type) *p, const char* json, size_t len);\
-MODEL_API int parse_##type##_array(array(type) *a, const char* json, size_t len);\
-MODEL_API int json_from_##type(const ptr(type) v, char *buf, size_t maxlen, size_t *len);
+MODEL_API int parse_##type##_array(array(type) *a, const char* json, size_t len); \
+MODEL_API char* type##_to_json(const ptr(type) v, int flags, size_t *len);
```
scareything, line 75: I'm guessing some flags will be added in the future? What are you thinking about for this? I'll put in a vote for the ability to return "pretty" json or not - or even just json with no newlines added.

tests/model_tests.cpp
```diff
@@ -164,11 +162,10 @@ TEST_CASE("test skipped fields") {
     REQUIRE(foo.barp->isOK);
     REQUIRE_THAT(foo.barp->msg, Catch::Matchers::Equals("hello world!"));
 
-    char json1[1024];
     size_t jsonlen;
-    CHECK(json_from_Foo(&foo, json1, sizeof(json1), &jsonlen) == 0);
+    char *json1 = Foo_to_json(&foo, 0, &jsonlen);
```
scareything, line 166: None of the tests/samples I see here demonstrate how to determine whether or not the encoding was successful. I'm assuming a NULL return value?

includes/ziti/model_support.h
ekoby: yes, that's the intent

scareything (approved):

scareything:
I think these changes are an improvement overall, but it would be nice if there was some way to not use the heap when encoding json. The latest tsdk encodes a model object to json every time it dials, and for example http/s services produce many dials.

includes/ziti/model_support.h
```diff
@@ -71,8 +73,10 @@ MODEL_API int cmp_##type(const type *lh, const type *rh); \
 MODEL_API void free_##type##_array(array(type) *ap);\
 MODEL_API int parse_##type(ptr(type) v, const char* json, size_t len);\
 MODEL_API int parse_##type##_ptr(ptr(type) *p, const char* json, size_t len);\
-MODEL_API int parse_##type##_array(array(type) *a, const char* json, size_t len);\
-MODEL_API int json_from_##type(const ptr(type) v, char *buf, size_t maxlen, size_t *len);
+MODEL_API int parse_##type##_array(array(type) *a, const char* json, size_t len); \
+/** write to fixeb buffer */                                 \
```
scareything, line 77: typo

PR #268, opened 2021-04-19
dump service configs

ekoby:


library/ziti.c
```diff
@@ -369,12 +369,18 @@ void ziti_dump(ziti_context ztx, int (*printer)(void *arg, const char *fmt, ...)
...
+    const char *cfg_json;
     MODEL_MAP_FOREACH(name, zs, &ztx->services) {
 
         printer(ctx, "%s: id[%s] perm(dial=%s,bind=%s)\n", zs->name, zs->id,
                 (zs->perm_flags & ZITI_CAN_DIAL ? "true" : "false"),
                 (zs->perm_flags & ZITI_CAN_BIND ? "true" : "false")
         );
+
+        MODEL_MAP_FOREACH(cfg, cfg_json, &zs->config) {
+            printer(ctx, "\tconfig[%s]=%s\n", cfg, cfg_json);
```
dovholuknf, line 382: this is probably fine for most stuff but if this is sensitive to an sdk user - would we want this or make it optional or provide some way to mask of 'sensitive data' at all?

dovholuknf (approved):
looks good other than the question of if it's good to always log it

library/ziti.c
ekoby: SDK does not log stuff directly, it is up to the app to decide where ziti dump ends up

PR #275, opened 2021-04-28
API versioning

ekoby:


deps/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ include(FetchContent)
 
 FetchContent_Declare(uv-mbed
         GIT_REPOSITORY https://github.com/netfoundry/uv-mbed.git
-        GIT_TAG v0.12.4
+        GIT_TAG v0.13.0
```
dovholuknf, line 7: why the bump to 13? it's api compatible isn't it?

dovholuknf (approved):

deps/CMakeLists.txt
ekoby: it changed behavior of um_http_init() wrt passed in URL

PR #276, opened 2021-05-04
Default posture queries

### LINKED ISSUE #274: implement default posture callbacks
* make Posture Query type enum

ekoby:
[fixes #274]

### DIFF
--- a/library/posture.c
+++ b/library/posture.c
@@ -19,13 +19,18 @@ limitations under the License.
 
 #if _WIN32
 #include <winnt.h>
+#include <wincrypt.h>
+#include <bcrypt.h>
+#include <tlhelp32.h>
+
+#pragma comment(lib, "netapi32.lib")
+#pragma comment(lib, "bcrypt.lib")
 
 // provided by libuv
 typedef NTSTATUS (NTAPI *sRtlGetVersion)
         (PRTL_OSVERSIONINFOW lpVersionInformation);
 extern sRtlGetVersion pRtlGetVersion;
 
-#pragma comment(lib, "netapi32.lib")
 extern DWORD NetGetJoinInformation (
         LPCWSTR               lpServer,
         LPWSTR                *lpNameBuffer,
@@ -74,7 +79,8 @@ static void ziti_pr_handle_domain(ziti_context ztx, const char *id, const char *
 
 static void ziti_pr_handle_os(ziti_context ztx, const char *id, const char *os_type, const char *os_version, const char *os_build);
 
-static void ziti_pr_handle_process(ziti_context ztx, char *id, char *path, bool is_running, char *sha_512_hash, char **signers, int num_signers);
+static void ziti_pr_handle_process(ziti_context ztx, const char *id, const char *path,
+                                   bool is_running, const char *sha_512_hash, char **signers, int num_signers);
 
 static void ziti_pr_send(ziti_context ztx);
 
@@ -87,6 +93,11 @@ static bool ziti_pr_is_info_errored(ziti_context ztx, char *id);
 static void default_pq_os(ziti_context ztx, const char *id, ziti_pr_os_cb response_cb);
 static void default_pq_mac(ziti_context ztx, const char *id, ziti_pr_mac_cb response_cb);
 static void default_pq_domain(ziti_context ztx, const char* id, ziti_pr_domain_cb cb);
+static void default_pq_process(ziti_context ztx, const char *id, const char *path, ziti_pr_process_cb cb);
+
+static char** get_signers(const char *path, int *signers_count);
+static int hash_sha512(uv_loop_t *loop, const char *path, unsigned char **out_buf, size_t *out_len);
+static bool check_running(const char *path);
 
 static void ziti_pr_free_pr_info_members(pr_info *info) {
     FREE(info->id)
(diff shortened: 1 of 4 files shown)


library/posture.c
```diff
@@ -254,8 +289,9 @@ void ziti_send_posture_data(ziti_context ztx) {
         if (ztx->opts->pq_domain_cb != NULL) {
             ztx->opts->pq_domain_cb(ztx, domainInfo->query->id, ziti_pr_handle_domain);
         } else {
-            ZITI_LOG(DEBUG, "%s cb not set requested for: service %s, policy: %s, check: %s", PC_DOMAIN_TYPE,
+            ZITI_LOG(VERBOSE, "using default %s cb for: service %s, policy: %s, check: %s", PC_DOMAIN_TYPE,
```
dovholuknf, line 292: why this one to verbose when the others are left at debug?

ekoby: i'll change all to verbose

library/posture.c
```diff
@@ -545,3 +585,273 @@ static void ziti_pr_handle_process(ziti_context ztx, char *id, char *path, bool
...
+    switch (os_info.wProductType) {
+        case 1: os = "windows";
+            break;
+        case 2:
+        case 3: os = "windowsserver";
+            break;
+        default:
+            os = "<unknown windows type>";
+    }
+    char winver[16];
```
dovholuknf, line 613: while this definitely seems like enough why not just make it bigger just because... it's short lived anyway. i worry :)

ekoby: 32?

ekoby: it won't crash, even if  the buffer is too short

library/posture.c
```diff
@@ -545,3 +585,273 @@ static void ziti_pr_handle_process(ziti_context ztx, char *id, char *path, bool
...
+
+static void default_pq_domain(ziti_context ztx, const char* id, ziti_pr_domain_cb cb) {
+#if _WIN32
+    uint32_t status;
+    LPWSTR buf;
+    NetGetJoinInformation(NULL, &buf, &status);
+    char domain[256];
+    sprintf_s(domain, sizeof(domain), "%ls", buf);
+    cb(ztx, id, domain);
+    NetApiBufferFree(buf);
```
dovholuknf, line 661: doesn't domain need to be free'd too?

ekoby: no, domain is on stack

library/posture.c
dovholuknf: i just worry that it'll be a hard to track down bug is all. each of those fields "can be" up to like 10 characters long as a string, right? if MS makes them crazy long some day on purpose or by mistake it'd be a problem. 32 seems like it'd be enough but why not just go ham and use 64?

library/posture.c
```diff
@@ -545,3 +585,273 @@ static void ziti_pr_handle_process(ziti_context ztx, char *id, char *path, bool
...
+                }
+                free(ex.ptr);
+            }
+        }
+    }
+
+#else
+    uv_utsname_t uname;
+    uv_os_uname(&uname);
+    ZITI_LOG(WARN, "not implemented on %s", uname.sysname);
```
dovholuknf, line 800: better log message maybe? "cannot check for process running on %s. not implemented" ?

library/posture.c
```diff
@@ -545,3 +585,273 @@ static void ziti_pr_handle_process(ziti_context ztx, char *id, char *path, bool
...
+
+    return rc;
+}
+
+static bool check_running(uv_loop_t *loop, const char *path) {
+    bool result = false;
+#if _WIN32
+    HANDLE sh = CreateToolhelp32Snapshot(TH32CS_SNAPPROCESS, 0);
+    if (sh == NULL) {
+        ZITI_LOG(ERROR, "failed: %d", GetLastError());
```
dovholuknf, line 754: why is the rest of the block allowed to continue if this fails? will line 763 still succeed if if `sh` is NULL?

library/posture.c
ekoby: log message shows function name

dovholuknf (approved):

library/posture.c
```diff
@@ -795,6 +797,21 @@ static bool check_running(uv_loop_t *loop, const char *path) {
...
+#elif __APPLE__ && __MACH__
+    int n_pids = proc_listallpids(NULL, 0);
+    unsigned long pids_sz = sizeof(pid_t) * (unsigned long)n_pids;
+    pid_t * pids = calloc(1, pids_sz);
+    proc_listallpids(pids, (int)pids_sz);
+    char proc_path[PROC_PIDPATHINFO_MAXSIZE];
+    for (int i=0; i < n_pids; i++) {
+        if (pids[i] == 0) continue;
+        proc_pidpath(pids[i], proc_path, sizeof(proc_path)); // returns strlen(proc_path)
+        if (strncasecmp(proc_path, path, sizeof(proc_path)) == 0) {
```
ekoby, line 809: I don't believe it should be case-insensitive

scareything: Why not? the default file systems on macOS are case-insensitive:
```
$ uname -s
Darwin
$ touch Foo
$ touch foo
$ ls ?oo
Foo
```

ekoby: :exploding_head: I did not know that

PR #282, opened 2021-05-14
use proper header for windows x86

ekoby:


dovholuknf (approved):
"certified fresh" - x86 builds now using -G "Visual Studio 16 2019". thanks!

PR #283, opened 2021-05-26
re-auths during MFA if API session became invalid

andrewpmartinez:
- add support for MFA in ziti-prox-c

### DIFF
--- a/programs/ziti-prox-c/proxy.c
+++ b/programs/ziti-prox-c/proxy.c
@@ -508,14 +508,34 @@ struct mfa_work {
 
 void mfa_response_cb(ziti_context ztx, void *mfa_ctx, int status, void *ctx);
 
+void prompt_stdin(char *buffer, size_t buflen) {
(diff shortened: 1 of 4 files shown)


programs/ziti-prox-c/proxy.c
```diff
@@ -503,23 +490,78 @@ static void on_ziti_event(ziti_context ztx, const ziti_event_t *event) {
...
+};
+
+void mfa_response_cb(ziti_context ztx, void *mfa_ctx, int status, void *ctx);
+
+void mfa_prompt(struct mfa_work *mfa_wr) {
+    uv_sleep(250);
+    char code[512];
+    printf("\nPlease provide your current MFA token: ");
+    fflush(stdout);
+    gets(code);
```
ekoby, line 516: `gets` shall not be used

andrewpmartinez: C makes me sad. I will use fgets.

ekoby (approved):

PR #286, opened 2021-06-08
Identity app data

ekoby:
~waiting for fix https://github.com/openziti/edge/pull/683/files to test~


inc_internal/internal_model.h
```diff
@@ -21,6 +21,15 @@ limitations under the License.
 #include "ziti/model_support.h"
 #include "ziti/ziti_model.h"
 
+#define ZITI_IDENTITY_DATA_MODEL(XX, ...) \
```
scareything, line 24: I'm curious why you added a new model instead of adding to the existing identity model?

scareything (approved):

PR #292, opened 2021-07-07
adds MFA option support

andrewpmartinez:
- endpoint wake/unlock state change
- MFA events instead of callbacks

### DIFF
--- a/library/posture.c
+++ b/library/posture.c
@@ -731,6 +731,22 @@ static void process_check_work(uv_work_t *w) {
     pcw->signers = get_signers(path, &pcw->num_signers);
 }
 
+void ziti_endpoint_state_change(ziti_context ztx, bool woken, bool unlocked) {
+    if(woken || unlocked) {
+        ziti_pr_endpoint_state_req state_req = {
+                .typeId = (char *) PC_ENDPOINT_STATE_TYPE,
+                .unlocked = unlocked,
+                .woken = woken
+        };
+
+        size_t obj_len;
+
+        char *obj = ziti_pr_endpoint_state_req_to_json(&state_req, 0, &obj_len);
+
+        ziti_pr_post(&ztx->controller, obj, obj_len, NULL, NULL);
+    }
+}
+
 static int hash_sha512(uv_loop_t *loop, const char *path, unsigned char **out_buf, size_t *out_len) {
     size_t digest_size = crypto_hash_sha512_bytes();
     unsigned char *digest = NULL;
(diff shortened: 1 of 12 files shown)


library/posture.c
```diff
@@ -731,6 +731,31 @@ static void process_check_work(uv_work_t *w) {
...
+        ziti_pr_endpoint_state_req state_req = {
+                .id = "0",
+                .typeId = (char *) PC_ENDPOINT_STATE_TYPE,
+                .unlocked = unlocked,
+                .woken = woken
+        };
+
+        size_t obj_len;
+
+        char *obj = ziti_pr_endpoint_state_req_to_json(&state_req, 0, &obj_len);
```
ekoby, line 752: when is this freed?

andrewpmartinez: `ziti_pr_post` calls `um_http_req_data` with `obj` as the `body`. The `um_http_req_data` should free it.

programs/ziti-prox-c/proxy.c
```diff
@@ -83,8 +83,9 @@ struct client {
 };
 
 static int process_args(int argc, char *argv[]);
+void mfa_auth_event_handler(ziti_context ztx, ziti_auth_query_mfa *aq_mfa);
 
-int main(int argc, char *argv[]) {
+        int main(int argc, char *argv[]) {
```
ekoby, line 88: indent?

programs/ziti-prox-c/proxy.c
```diff
@@ -505,11 +510,10 @@ struct mfa_work {
     uv_work_t w;
     ziti_context ztx;
     ziti_auth_query_mfa *aq_mfa;
-    ziti_ar_mfa_cb response_cb;
-    void *mfa_ctx;
+    //ziti_mfa_cb response_cb;
```
ekoby, line 513: cleanup

programs/ziti-prox-c/proxy.c
andrewpmartinez: fixed

programs/ziti-prox-c/proxy.c
andrewpmartinez: fixed

library/posture.c
ekoby: ok seems that controller(ziti_ctrl.c) frees is (free_body_cb). It is ok, but the preferred way is to free memory in the same layer where it was allocated

ekoby (approved):

dovholuknf (approved):

PR #298, opened 2021-07-21
fix linux OS check; cleanup MAC list

ekoby:


library/posture.c
```diff
@@ -626,32 +626,51 @@ static void default_pq_os(ziti_context ztx, const char *id, ziti_pr_os_cb respon
...
-    build = uname.release;
+    ver = uname.release;
+    build = uname.version;
 #endif
 
     response_cb(ztx, id, os, ver, build);
 }
 
+static bool non_zero_addr(const char* addr, int addr_size) {
+    for (int i = 0; i < addr_size; i++) {
```
dovholuknf, line 637: what if it's only whitespace?

ekoby: it is not text, addr in binary

dovholuknf (approved):

PR #299, opened 2021-07-28
timeoutRemaining defaults to -1 for controllers that do not supply it

andrewpmartinez:


includes/ziti/ziti_model.h
```diff
@@ -64,7 +64,7 @@ XX(query_type, string, none, queryType, __VA_ARGS__) \
 XX(process, ziti_process, ptr, process, __VA_ARGS__) \
 XX(processes, ziti_process, array, processes, __VA_ARGS__) \
 XX(timeout, int, none, timeout, __VA_ARGS__) \
-XX(timeoutRemaining, int, none, timeoutRemaining, __VA_ARGS__)
+XX(timeoutRemaining, int, ptr, timeoutRemaining, __VA_ARGS__)
```
mary-dcouto, line 67: do we need to make this int field a pointer? Is there no other way to fix this?

andrewpmartinez: It has to be a pointer.

This model is used by hosting applications and to parse the REST responses from the controller.

The only way to detect if the controller didn't send the value is to use a pointer - otherwise, it will be initialized to 0 and we can't tell if the controller didn't send the value vs the controller sending a 0.

mary-dcouto (approved):

PR #302, opened 2021-08-17
Ztx enable disable shutdown

### LINKED ISSUE #110: issuing ziti_shutdown on windows produces Assertion failure
Steps to reproduce:
* connect an identity 
* verify things work properly
* try to reclaim memory/remove identity using `ziti_shutdown`
* observe: 
```
Assertion failed!

Program: c:\git\github\openziti\ziti-tunnel-win\service\ziti-tunnel.exe
File: ../deps/ziti-sdk-c/deps/uv-mbed/deps/libuv/src/win/async.c, Line 76

Expression: !(handle->flags & UV_HANDLE_CLOSING)
```

### LINKED ISSUE #301: Implement disabling/enabling of Ziti context
Here is the things that happen on disable:

- active `ziti_connection`s are disconnected, appropriate callback receives `ZITI_DISABLED` error code
- `ZitiServiceEvent` is sent with all current services removed
- connections to Edge Routers are closed, with corresponding `ZitiEdgeRouteEvent`s
- ziti_context terminates API session (controller /logout), with corresponding `ZitiContextEvent`
- all model data (services, sessions, api session) is freed

if context is disabled it can be re-enabled/restarted by calling `ziti_set_enabled(ztx, true)`. Re-enabling puts it back into 
running state

ekoby:
[fixes #301]
[fixes #110]

### DIFF
--- a/includes/ziti/error_defs.h
+++ b/includes/ziti/error_defs.h
@@ -87,7 +87,9 @@ limitations under the License.
     /** attempted to verify or retrieve details of an MFA enrollment that has not been completed */ \
     XX(MFA_NOT_ENROLLED, "the current identity has not completed MFA enrollment") \
     /** not found, usually indicates stale reference or permission */    \
-    XX(NOT_FOUND, "entity no longer exists or is no longer accessible") \
+    XX(NOT_FOUND, "entity no longer exists or is no longer accessible")  \
+    /** operation attempted while ziti_context is not enabled */                    \
+    XX(DISABLED, "ziti context is disabled") \
     /** Inspired by the Android SDK: What a Terrible Failure. A condition that should never happen. */ \
     XX(WTF, "WTF: programming error")
 
--- a/library/connect.c
+++ b/library/connect.c
@@ -1276,8 +1276,12 @@ static void process_edge_message(struct ziti_conn *conn, message *msg, int code)
 
     if (msg == NULL) {
         if (conn->state == Bound) {
-            CONN_LOG(DEBUG, "binding lost due to failed channel [%d/%s]", code, ziti_errorstr(code));
-            ziti_rebind(conn);
+            if (code == ZITI_DISABLED) {
+                conn->client_cb(conn, NULL, ZITI_DISABLED, NULL);
+            } else {
+                CONN_LOG(DEBUG, "binding lost due to failed channel [%d/%s]", code, ziti_errorstr(code));
+                ziti_rebind(conn);
+            }
             return;
         }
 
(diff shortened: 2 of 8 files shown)


library/ziti.c
```diff
@@ -206,14 +212,72 @@ int ziti_init_opts(ziti_options *options, uv_loop_t *loop) {
     return ZITI_OK;
 }
 
-static void ziti_init_async(uv_async_t *ar) {
-    struct ziti_init_req *init_req = ar->data;
+extern bool ziti_is_enabled(ziti_context ztx) {
+    return ztx->enabled;
```
dovholuknf, line 216: should this guard against a uninitialized/null'ed context? it'd be friendlier to not crash imo... but maybe that's just me? :)

ekoby: just you

otherwise we need to do it everywhere

includes/ziti/error_defs.h
```diff
@@ -87,7 +87,9 @@ limitations under the License.
     /** attempted to verify or retrieve details of an MFA enrollment that has not been completed */ \
     XX(MFA_NOT_ENROLLED, "the current identity has not completed MFA enrollment") \
     /** not found, usually indicates stale reference or permission */    \
-    XX(NOT_FOUND, "entity no longer exists or is no longer accessible") \
+    XX(NOT_FOUND, "entity no longer exists or is no longer accessible")  \
+    /** operation attempted while ziti_context is not enabled */                    \
+    XX(DISABLED, "ziti context is disabled") \
```
dovholuknf, line 92: i assume `ZITI_OK` is the antithesis of `DISABLED`?

ekoby: ZITI_OK is general _no error_ result

library/connect.c
```diff
@@ -488,6 +488,8 @@ static void ziti_connect_async(uv_async_t *ar) {
 }
 
 static int do_ziti_dial(ziti_connection conn, const char *service, ziti_dial_opts *dial_opts, ziti_conn_cb conn_cb, ziti_data_cb data_cb) {
+    if (!conn->ziti_ctx->enabled) return ZITI_DISABLED;
```
dovholuknf, line 491: why not use `ziti_is_enabled` here? same comment below

library/ziti.c
```diff
@@ -206,14 +212,72 @@ int ziti_init_opts(ziti_options *options, uv_loop_t *loop) {
...
+
+        ev.type = ZitiContextEvent;
+        ev.event.ctx.ctrl_status = ZITI_DISABLED;
+
+        ziti_send_event(ztx, &ev);
+    }
+}
+
+static void ziti_start_internal(ziti_context ztx, void *init_req) {
+    if (!ztx->enabled) {
```
dovholuknf, line 269: personally, i think this could use some log statements. this might be hard to debug/diagnose without.

dovholuknf (approved):

PR #304, opened 2021-08-19
ziti_identity should always be available after the first login

ekoby:
make ziti_get_identity() return valid value even after session is expired

### DIFF
--- a/library/ziti.c
+++ b/library/ziti.c
@@ -340,7 +340,15 @@ const ziti_version *ziti_get_controller_version(ziti_context ztx) {
 }
 
 const ziti_identity *ziti_get_identity(ziti_context ztx) {
-    return ztx->session ? ztx->session->identity : NULL;
+    if (ztx->identity_data) {
+        return (const ziti_identity *) ztx->identity_data;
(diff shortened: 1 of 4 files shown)


library/ziti.c
```diff
@@ -435,12 +431,17 @@ void ziti_dump(ziti_context ztx, int (*printer)(void *arg, const char *fmt, ...)
...
+    } else {
+        printer(ctx, "unknown - never logged in\n");
+    }
 
     printer(ctx, "\n=================\nSession:\n");
 
     if (ztx->session) {
-        printer(ctx, "Session Info: id[%s] name[%s] api_session[%s]\n",
-                ztx->session->identity->id, ztx->session->identity->name, ztx->session->id);
+        printer(ctx, "Session Info: api_session[%s]\n", ztx->session->id);
```
andrewpmartinez, line 444: Is this info available somewhere else?

andrewpmartinez (approved):

library/ziti.c
ekoby: right above

PR #305, opened 2021-08-23
changes service update logic

andrewpmartinez:
swap from a deep compare on service definitions to deep compare on
configurations and updatedAt comparisons for the rest of the service.
Ignores posture query values.

### DIFF
--- a/includes/ziti/ziti_model.h
+++ b/includes/ziti/ziti_model.h
@@ -80,7 +80,8 @@ XX(permissions, string, array, permissions, __VA_ARGS__) \
 XX(encryption, bool, none, encryptionRequired, __VA_ARGS__) \
 XX(perm_flags, int, none, NULL, __VA_ARGS__) \
 XX(config, json, map, config, __VA_ARGS__) \
-XX(posture_query_set, ziti_posture_query_set, array, postureQueries, __VA_ARGS__)
+XX(posture_query_set, ziti_posture_query_set, array, postureQueries, __VA_ARGS__) \
+XX(updated_at,string, none, updatedAt, __VA_ARGS__)
 
 #define ZITI_CLIENT_CFG_V1_MODEL(XX, ...) \
 XX(hostname, string, none, hostname, __VA_ARGS__) \
--- a/includes/ziti/ziti_model.h
+++ b/includes/ziti/ziti_model.h
@@ -81,6 +81,7 @@ XX(encryption, bool, none, encryptionRequired, __VA_ARGS__) \
 XX(perm_flags, int, none, NULL, __VA_ARGS__) \
 XX(config, json, map, config, __VA_ARGS__) \
 XX(posture_query_set, ziti_posture_query_set, array, postureQueries, __VA_ARGS__) \
+XX(posture_query_map, ziti_posture_query_set, map, posturePolicies, __VA_ARGS__) \
 XX(updated_at,string, none, updatedAt, __VA_ARGS__)
 
 #define ZITI_CLIENT_CFG_V1_MODEL(XX, ...) \
--- a/library/posture.c
+++ b/library/posture.c
@@ -93,11 +93,14 @@ static void ziti_pr_send_individually(ziti_context ztx);
 static bool ziti_pr_is_info_errored(ziti_context ztx, const char *id);
 
 static void default_pq_os(ziti_context ztx, const char *id, ziti_pr_os_cb response_cb);
+
 static void default_pq_mac(ziti_context ztx, const char *id, ziti_pr_mac_cb response_cb);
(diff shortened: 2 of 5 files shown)


library/ziti.c
```diff
@@ -643,23 +643,35 @@ static void ziti_re_auth(ziti_context ztx) {
...
 
+// is_service_updated returns 0 if the direct service properties
+// and configurations have not been altered. Will return non-0
+// values if they have. This ignores posture query alterations.
+static int is_service_updated(ziti_service *new, ziti_service *old) {
+    if (strcmp(old->updated_at, new->updated_at) != 0) {
+        return 1;
+    }
+
+    return model_map_compare(&old->config, &new->config, get_model_map_meta());
```
ekoby, line 672: pretty sure it is the wrong type_meta to pass in here. It should the type_meta of the values in the map

library/ziti.c
```diff
@@ -643,23 +643,35 @@ static void ziti_re_auth(ziti_context ztx) {
...
 }
 
+// is_service_updated returns 0 if the direct service properties
+// and configurations have not been altered. Will return non-0
+// values if they have. This ignores posture query alterations.
+static int is_service_updated(ziti_service *new, ziti_service *old) {
+    if (strcmp(old->updated_at, new->updated_at) != 0) {
+        return 1;
+    }
+
```
ekoby, line 671: didn't we discuss that it would also trigger on posture passing state change?

library/ziti.c
andrewpmartinez: Sure is. Fixed.

library/ziti.c
andrewpmartinez: Sure did, fixed.

includes/ziti/ziti_model.h
```diff
@@ -80,7 +80,9 @@ XX(permissions, string, array, permissions, __VA_ARGS__) \
 XX(encryption, bool, none, encryptionRequired, __VA_ARGS__) \
 XX(perm_flags, int, none, NULL, __VA_ARGS__) \
 XX(config, json, map, config, __VA_ARGS__) \
-XX(posture_query_set, ziti_posture_query_set, array, postureQueries, __VA_ARGS__)
+XX(posture_query_set, ziti_posture_query_set, array, postureQueries, __VA_ARGS__) \
+XX(posture_query_map, ziti_posture_query_set, map, posturePolicies, __VA_ARGS__) \
```
andrewpmartinez, line 84: Adds map version of posture sets.

library/posture.c
```diff
@@ -236,39 +238,36 @@ void ziti_send_posture_data(ziti_context ztx) {
...
-            setIdx++;
         }
     }
 
-    // mark all responses obsolete in case they were removed
+    // mark responses obsolete in case they were removed
     pr_info *resp;
     MODEL_MAP_FOREACH(name, resp, &ztx->posture_checks->responses) {
-        resp->obsolete = true;
+        if (!resp->pending && !resp->should_send) {
```
andrewpmartinez, line 268: This was marking process responses as obsolete before they were processed on the worker. The worker would finish and the process response would not be sent as it was marked obsolete.

library/posture.c
```diff
@@ -327,15 +326,14 @@ void ziti_send_posture_data(ziti_context ztx) {
         }
     }
 
-    if (procInfo != NULL) {
+    if (model_map_size(&processes) > 0) {
```
andrewpmartinez, line 329: `procInfo` would be null on subsequent runs. Causing process responses to not be submitted. We have a map of process responses now, so we can simply check that.

library/ziti.c
```diff
@@ -461,6 +468,16 @@ void ziti_dump(ziti_context ztx, int (*printer)(void *arg, const char *fmt, ...)
         MODEL_MAP_FOREACH(cfg, cfg_json, &zs->config) {
             printer(ctx, "\tconfig[%s]=%s\n", cfg, cfg_json);
         }
+
```
andrewpmartinez, line 471: Added posture queries to `ziti_dump`

library/ziti.c
```diff
@@ -882,9 +990,25 @@ static void edge_routers_cb(ziti_edge_router_array ers, const ziti_error *err, v
...
+        ztx->identity_data = data;
+    }
+
+    update_ctrl_status(ztx, FIELD_OR_ELSE(err, err, 0), FIELD_OR_NULL(err, message));
+}
+
+static void session_post_auth_query_cb(ziti_context ztx, int status, void *ctx) {
     ziti_session *session = ztx->session;
 
+    ziti_ctrl_current_identity(&ztx->controller, update_identity_data, ztx);
```
andrewpmartinez, line 1010: This was moved to after MFA is completed. It was 401'ing all the time on MFA identities.

ekoby (approved):

PR #307, opened 2021-09-01
use ZTX_LOG in more places, early returns on service update checking

andrewpmartinez:
- replaces ZITI_LOG  calls with ZTX_LOG where possible
- service update checking will now return earlier rather than performing
  additional checks

### DIFF
--- a/includes/ziti/ziti_log.h
+++ b/includes/ziti/ziti_log.h
@@ -43,8 +43,6 @@ enum DebugLevel {
 if (level <= ziti_log_level()) { ziti_logger(level, __FILE__, __LINE__, __func__, fmt, ##__VA_ARGS__); }\
 } while(0)
 
-#define ZTX_LOG(lvl, fmt, ...) ZITI_LOG(lvl, "ztx[%d] " fmt, ztx->id, ##__VA_ARGS__)
-
 #ifdef __cplusplus
 extern "C" {
 #endif
--- a/includes/ziti/ziti_log.h
+++ b/includes/ziti/ziti_log.h
@@ -43,6 +43,8 @@ enum DebugLevel {
 if (level <= ziti_log_level()) { ziti_logger(level, __FILE__, __LINE__, __func__, fmt, ##__VA_ARGS__); }\
 } while(0)
 
+#define ZTX_LOG(lvl, fmt, ...) ZITI_LOG(lvl, "ztx[%d] " fmt, ztx->id, ##__VA_ARGS__)
+
 #ifdef __cplusplus
 extern "C" {
 #endif
--- a/library/auth_queries.c
+++ b/library/auth_queries.c
@@ -316,13 +316,13 @@ void ziti_mfa_auth_internal_cb(void *empty, const ziti_error *err, void *ctx) {
         }
     } else {
         if (mfa_auth_ctx->ztx->auth_queries->outstanding_auth_query_ctx != NULL) {
-            ziti_mfa_auth_ctx *ctx = mfa_auth_ctx->ztx->auth_queries->outstanding_auth_query_ctx;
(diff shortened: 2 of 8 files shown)


includes/ziti/ziti_log.h
```diff
@@ -43,6 +43,8 @@ enum DebugLevel {
 if (level <= ziti_log_level()) { ziti_logger(level, __FILE__, __LINE__, __func__, fmt, ##__VA_ARGS__); }\
 } while(0)
 
+#define ZTX_LOG(lvl, fmt, ...) ZITI_LOG(lvl, "ztx[%d] " fmt, ztx->id, ##__VA_ARGS__)
```
dovholuknf, line 46: do we really want this exposed to consumers?

library/auth_queries.c
```diff
@@ -159,29 +159,31 @@ void ziti_auth_query_process(ziti_context ztx, void(*cb)(ziti_context, int, void
...
-        ZITI_LOG(ERROR, "error during create MFA call: %d - %s - %s", err->http_code, err->code, err->message);
+        ZTX_LOG(ERROR, "error during create MFA call: %d - %s - %s", err->http_code, err->code, err->message);
         mfa_enroll_cb_ctx->cb(mfa_enroll_cb_ctx->ztx, err->err, NULL, mfa_enroll_cb_ctx->cb_ctx);
         FREE(ctx);
     }
 }
 
 void ziti_mfa_enroll_get_internal_cb(ziti_mfa_enrollment *mfa_enrollment, const ziti_error *err, void *ctx) {
     ziti_mfa_enroll_cb_ctx *mfa_enroll_cb_ctx = ctx;
+    ziti_context ztx = mfa_enroll_cb_ctx->ztx;
```
dovholuknf, line 182: same comment here - is ztx used somehow?

library/auth_queries.c
```diff
@@ -159,29 +159,31 @@ void ziti_auth_query_process(ziti_context ztx, void(*cb)(ziti_context, int, void
...
 
-    ZITI_LOG(INFO, "sending auth query mfa authentication event, authentication cannot proceed if not handled");
+    ZTX_LOG(INFO, "sending auth query mfa authentication event, authentication cannot proceed if not handled");
     ziti_auth_query_mfa_process(mfa_auth_ctx);
 }
 
 
 void ziti_mfa_enroll_post_internal_cb(void *empty, const ziti_error *err, void *ctx) {
     ziti_mfa_enroll_cb_ctx *mfa_enroll_cb_ctx = ctx;
+    ziti_context ztx = mfa_enroll_cb_ctx->ztx;
```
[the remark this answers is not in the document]
andrewpmartinez: `ZTX_LOG` macro expects a scope defined `ztx`

library/auth_queries.c
andrewpmartinez: ZTX_LOG macro expects a scope defined ztx

library/auth_queries.c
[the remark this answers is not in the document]
dovholuknf: ah. well then that makes a lot more sense :) thx

includes/ziti/ziti_log.h
andrewpmartinez: Ahh hmm. I don't know what would happen if they tried to use it. 

I can move it to `zt_internal.h`. Is that a better place for it @ekoby ?

ekoby: yes

programs/ziti-prox-c/proxy.c
```diff
@@ -479,6 +479,24 @@ static void on_ziti_event(ziti_context ztx, const ziti_event_t *event) {
                     service_check_cb(ztx, *sp, ZITI_OK, app_ctx);
                 }
             }
+
+            if (event->event.service.changed != NULL) {
```
andrewpmartinez, line 483: Allows for mfa prompting on posture check timeouts.

dovholuknf (approved):

PR #309, opened 2021-09-14
adds api_session_state to internal ztx

andrewpmartinez:
adds api_session_state to internal ztx

    - internal change only
    - api_session_state is an enum of possible authentication states
    - states include unauthenticated, authenticated, and partially
      authenticated
    - partial authenticated represents the state where auth queries are
      being provided (i.e. mfa)
    - update service and channel reconnects now wait for full authentication
    - re-auth paths now reset authenticate state to unauthenticated
    - the current api session is now updated after authentication if auth
      queries were present
    - adds partial authentication status ctrl update
    - adds api_session_state for 'authenticating'
    - adds debug output
    - adds guards against re-auths during auth
    - renames ziti_session to ziti_api_session
    - checks for api session expiration on re-auth requests
    - adds api_session_state to ziti_dump
    - adds check status from auth queries
    - add more status updates


inc_internal/internal_model.h
```diff
@@ -148,7 +148,7 @@ DECLARE_MODEL(ziti_edge_router, ZITI_EDGE_ROUTER_MODEL)
 
 DECLARE_MODEL(ziti_net_session, ZITI_NET_SESSION_MODEL)
 
-DECLARE_MODEL(ziti_session, ZITI_SESSION_MODEL)
+DECLARE_MODEL(ziti_session, ZITI_API_SESSION_MODEL)
```
ekoby, line 151: any reason to rename the model but not the type?

library/ziti.c
```diff
@@ -440,8 +459,8 @@ void ziti_dump(ziti_context ztx, int (*printer)(void *arg, const char *fmt, ...)
 
     printer(ctx, "\n=================\nSession:\n");
 
-    if (ztx->session) {
-        printer(ctx, "Session Info: api_session[%s]\n", ztx->session->id);
+    if (ztx->api_session) {
+        printer(ctx, "Session Info: api_session[%s]\n", ztx->api_session->id);
```
ekoby, line 463: add new auth state to ziti_dump

library/ziti.c
```diff
@@ -638,19 +656,43 @@ void ziti_force_session_refresh(ziti_context ztx) {
 
 void ziti_re_auth_with_cb(ziti_context ztx, void(*cb)(ziti_session *, const ziti_error *, void *), void *ctx) {
     ZTX_LOG(WARN, "starting to re-auth with ctlr[%s]", ztx->opts->controller);
+
+    //to force a re-auth, set api_session_sate = ZitiApiSessionStateUnauthenticated
+    if (ztx->api_session_state == ZitiApiSessionStatePartiallyAuthenticated || ztx->api_session_state == ZitiApiSessionStateAuthStarted) {
```
ekoby, line 661: do we need to check if api_session has expired and ztx need to restart auth from very begining?

library/ziti.c
```diff
@@ -992,11 +1033,31 @@ static void update_identity_data(ziti_identity_data *data, const ziti_error *err
...
+        ziti_set_api_session(ztx, session);
+        ZTX_LOG(VERBOSE, "api session after authentication query processing complete");
+    }
+}
+
 static void session_post_auth_query_cb(ziti_context ztx, int status, void *ctx) {
-    ziti_session *session = ztx->session;
+    ziti_session *session = ztx->api_session;
+
+    ziti_set_fully_authenticated(ztx);
```
ekoby, line 1052: should this depend on `status`? what if authcode was not accepted?

library/ziti.c
```diff
@@ -1052,23 +1118,26 @@ static void session_cb(ziti_session *session, const ziti_error *err, void *ctx)
...
     if (session) {
-        ZTX_LOG(DEBUG, "%s successfully => api_session[%s]", ztx->session ? "refreshed" : "logged in", session->id);
+        ZTX_LOG(DEBUG, "%s successfully => api_session[%s]", ztx->api_session ? "refreshed" : "logged in", session->id);
 
-        set_session(ztx, session);
+        ziti_set_api_session(ztx, session);
 
         ziti_auth_query_init(ztx);
 
+        ztx->api_session_state = ZitiApiSessionStatePartiallyAuthenticated;
```
ekoby, line 1127: if I read this correctly `ziti_set_api_session()` may set state to fully auth'ed if there are no auth_queries, 
but then this line will set it back to partial auth'ed

is this correct?

ekoby: looks like it is already done there

library/ziti.c
```diff
@@ -220,11 +220,31 @@ extern void ziti_set_enabled(ziti_context ztx, bool enabled) {
     ziti_queue_work(ztx, enabled ? ziti_start_internal : ziti_stop_internal, NULL);
 }
 
+void ziti_set_auth_started(ziti_context ztx) {
+    free_ziti_session(ztx->api_session);
+    FREE(ztx->api_session);
+    ztx->api_session_state = ZitiApiSessionStateAuthStarted;
+}
+
```
ekoby, line 228: probably worth logging these state transitions

inc_internal/internal_model.h
andrewpmartinez: Rename completed.

library/ziti.c
andrewpmartinez: Yea I've enhanced this logic to check for that condition.

library/ziti.c
andrewpmartinez: Fixed

library/ziti.c
andrewpmartinez: fixed

library/ziti.c
andrewpmartinez: It should looking at the code. 

When a non-ZITI_OK status is provided what should we do? Right now I set it back to unauthenticated, but if we have auth queries that don't pause on user input, this will repeat on the session_refresh timer.

library/ziti.c
andrewpmartinez: This was changed. It is now a single call to `ziti_set_api_session` which in turn uses the state transition functions which provide logging. `ziti_set_api_session` detect partial/full auth based on outstanding auth queries.

library/ziti.c
```diff
@@ -992,83 +1063,119 @@ static void update_identity_data(ziti_identity_data *data, const ziti_error *err
...
+        ziti_posture_init(ztx, 20);
+
+        if (!ztx->no_current_edge_routers) {
+            ziti_ctrl_current_edge_routers(&ztx->controller, edge_routers_cb, ztx);
+        }
+    } else {
+        ZTX_LOG(VERBOSE, "transitioning to unauthenticated, unhandled status[%s]", ziti_errorstr(status));
+        ziti_set_unauthenticated(ztx); //disable?
+        update_ctrl_status(ztx, ZITI_NOT_AUTHORIZED, NULL);
```
andrewpmartinez, line 1133: I added a few more of these in related branches. I don't know if this matters or not.

ekoby (approved):

PR #311, opened 2021-09-20
adds defensive channel reconnect after api session change

andrewpmartinez:
Previously it was possible for channels to remain connected to Edge Routers with previous API Sessions. This would occur in scenarios where the SDK requested a new API Session but did not invalidate the previous one. Under normal conditions, this shouldn't happen, but bugs have caused it to happen in the past. It can also happen when ERs are not properly receiving updates from controllers and C-SDK channels remain connected when all the others have disconnected.

### DIFF
--- a/library/channel.c
+++ b/library/channel.c
@@ -252,7 +252,7 @@ bool ziti_channel_is_connected(ziti_channel_t *ch) {
 
 static void ziti_channel_set_api_session_id(ziti_channel_t *ch, const char* api_session_id) {
     memset(ch->api_session_id, '\0', sizeof(ch->api_session_id));
-    strncpy(ch->api_session_id, api_session_id, strlen(api_session_id));
+    strncpy(ch->api_session_id, api_session_id, strlen(ch->api_session_id));
 }
 
 static ziti_channel_t *new_ziti_channel(ziti_context ztx, const char *ch_name, const char *url) {
--- a/library/channel.c
+++ b/library/channel.c
@@ -250,6 +250,11 @@ bool ziti_channel_is_connected(ziti_channel_t *ch) {
     return ch->state == Connected;
 }
 
+static void ziti_channel_set_api_session_id(ziti_channel_t *ch, const char* api_session_id) {
+    memset(ch->api_session_id, '\0', sizeof(ch->api_session_id));
+    strncpy(ch->api_session_id, api_session_id, strlen(api_session_id));
+}
+
 static ziti_channel_t *new_ziti_channel(ziti_context ztx, const char *ch_name, const char *url) {
     ziti_channel_t *ch = calloc(1, sizeof(ziti_channel_t));
     ziti_channel_init(ztx, ch, channel_counter++, ztx->tlsCtx);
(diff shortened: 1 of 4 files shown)


library/ziti.c
```diff
@@ -1128,6 +1128,7 @@ static void session_post_auth_query_cb(ziti_context ztx, int status, void *ctx)
         if (!ztx->no_current_edge_routers) {
             ziti_ctrl_current_edge_routers(&ztx->controller, edge_routers_cb, ztx);
         }
+
```
andrewpmartinez, line 1131: This was tested by adding the following function above:

```
void debug_timer(uv_timer_t *t){
    ziti_context ztx = t->data;

    ziti_re_auth(ztx);

    FREE(t);
}
```


and starting a timer here

```
        NEWP(timer, uv_timer_t);
        uv_timer_init(ztx->loop, timer);
        timer->data = ztx;
        uv_timer_start(timer, debug_timer, 15000, 0)
```

library/channel.c
```diff
@@ -238,6 +265,10 @@ static ziti_channel_t *new_ziti_channel(ziti_context ztx, const char *ch_name, c
     int hostoffset = ingress.field_data[UF_HOST].off;
     snprintf(host, sizeof(host), "%*.*s", hostlen, hostlen, url + hostoffset);
 
+    //save the api_session_id that was used to request a session and create this channel
+    memset(ch->api_session_id, '\0', sizeof(ch->api_session_id));
+    strncpy(ch->api_session_id, ztx->api_session->id, strlen(ztx->api_session->id) - 1);
```
ekoby, line 270: why `-1`?

library/channel.c
```diff
@@ -846,6 +868,11 @@ static void on_channel_connect_internal(uv_connect_t *req, int status) {
             uv_mbed_t *mbed = (uv_mbed_t *) req->handle;
             uv_mbed_read(mbed, ziti_alloc_cb, on_channel_data);
             ch->reconnect_count = 0;
+
+            //save the api_session_id that was used to request a session and create this channel
+            memset(ch->api_session_id, '\0', sizeof(ch->api_session_id));
+            strncpy(ch->api_session_id, ch->ctx->api_session->id, strlen(ch->ctx->api_session->id) - 1);
```
ekoby, line 874: `-1` is most definitely wrong

ekoby: is it leftover from debugging?

library/channel.c
andrewpmartinez: Copy pasta! Centralized in a function.

library/channel.c
andrewpmartinez: The API session id is set during initialization and during connection. If debug output was shown while a channel before initial connect wanted it to be clear what the intended API session id was.

library/channel.c
```diff
@@ -250,6 +250,11 @@ bool ziti_channel_is_connected(ziti_channel_t *ch) {
     return ch->state == Connected;
 }
 
+static void ziti_channel_set_api_session_id(ziti_channel_t *ch, const char* api_session_id) {
+    memset(ch->api_session_id, '\0', sizeof(ch->api_session_id));
+    strncpy(ch->api_session_id, api_session_id, strlen(api_session_id));
```
ekoby, line 255: should be `strncpy(ch->api_session_id, api_session_id, sizeof(ch->api_session_id))` in case passed in value is longer than expected

andrewpmartinez: changed

PR #312, opened 2021-09-20
Instance.id.support

andrewpmartinez:
The Ziti controller has supported an instance id since 0.19. Which can be used to detect if the controller has been restarted. In this PR this is used to determine if all posture data should be resubmitted.

### DIFF
--- a/library/posture.c
+++ b/library/posture.c
@@ -130,7 +130,7 @@ void ziti_posture_init(ziti_context ztx, long interval_secs) {
 
         pc->timer = timer;
         pc->interval = (double) interval_secs;
-        pc->previous_session_id = NULL;
+        pc->previous_api_session_id = NULL;
         pc->must_send_every_time = true;
         pc->must_send = false;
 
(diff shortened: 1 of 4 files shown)


library/posture.c
```diff
@@ -183,13 +185,24 @@ void ziti_send_posture_data(ziti_context ztx) {
     ZTX_LOG(VERBOSE, "starting to send posture data");
     bool new_session_id = ztx->posture_checks->previous_session_id == NULL || strcmp(ztx->posture_checks->previous_session_id, ztx->api_session->id) != 0;
 
-    if (new_session_id || ztx->posture_checks->must_send_every_time) {
-        ZTX_LOG(DEBUG, "posture checks either never sent or api session changed, must_send = true");
+    bool new_controller_instance = (ztx->posture_checks->controller_instance_id == NULL && ztx->controller.instance_id != NULL) || strcmp(ztx->posture_checks->controller_instance_id, ztx->controller.instance_id) != 0;
```
dovholuknf, line 188: could you please output a log when this boolean becomes only  true ?

dovholuknf (approved):

library/posture.c
andrewpmartinez: done

library/ziti_ctrl.c
```diff
@@ -169,6 +170,13 @@ static void ctrl_resp_cb(um_http_resp_t *r, void *data) {
         }
 
         resp->new_address = find_header(r, "ziti-ctrl-address");
+
+        const char *instance_id = find_header(r, "ziti-instance-id");
+
+        if (instance_id) {
+            FREE(resp->ctrl->instance_id);
+            resp->ctrl->instance_id = strdup(instance_id);
+        }
```
andrewpmartinez, line 179: This section is the only part I actually changed. The rest is auto-formatting from CLion (default installation).

library/posture.c
```diff
@@ -181,15 +181,30 @@ void ziti_send_posture_data(ziti_context ztx) {
     }
 
     ZTX_LOG(VERBOSE, "starting to send posture data");
-    bool new_session_id = ztx->posture_checks->previous_session_id == NULL || strcmp(ztx->posture_checks->previous_session_id, ztx->api_session->id) != 0;
+    bool new_session_id = ztx->posture_checks->previous_api_session_id == NULL || strcmp(ztx->posture_checks->previous_api_session_id, ztx->api_session->id) != 0;
+
+    bool new_controller_instance = (ztx->posture_checks->controller_instance_id == NULL && ztx->controller.instance_id != NULL) || strcmp(ztx->posture_checks->controller_instance_id, ztx->controller.instance_id) != 0;
+
+    if(new_controller_instance){
+        ZTX_LOG(WARN, "a new controller instance was detected");
```
ekoby, line 189: I don't like `instance` since the controller does not actually change

I would change this message to "controller restart detected"

ekoby:
this will need to be reworked if we have distributed controller, right?

library/ziti_ctrl.c
andrewpmartinez: Removed formatting

PR #313, opened 2021-09-23
API Session Token Cleanup, API Session State Fixes

andrewpmartinez:


library/ziti.c
```diff
@@ -239,6 +239,18 @@ void ziti_set_unauthenticated(ziti_context ztx) {
     free_ziti_api_session(ztx->api_session);
     FREE(ztx->api_session);
     ztx->api_session_state = ZitiApiSessionStateUnauthenticated;
+
+    ctrl_clear_api_session(&ztx->controller);
```
andrewpmartinez, line 243: Clear API Session Token in controller client on 'unauthenticated' and below on 'impossible to authenticate'

library/ziti.c
```diff
@@ -1226,11 +1237,14 @@ static void api_session_cb(ziti_api_session *session, const ziti_error *err, voi
         if (errCode == ZITI_NOT_AUTHORIZED) {
             if (ztx->api_session || !init_req->start) {
                 ZTX_LOG(DEBUG, "received NOT_AUTHORIZED attempting re-auth");
+                ziti_set_unauthenticated(ztx);
```
andrewpmartinez, line 1240: missing api session state change

library/ziti.c
```diff
@@ -1226,11 +1237,14 @@ static void api_session_cb(ziti_api_session *session, const ziti_error *err, voi
...
                 ziti_re_auth(ztx);
                 errCode = ztx->ctrl_status; // do not trigger event yet
             } else {
                 // cannot login or re-auth -- identity no longer valid
                 // notify service removal, and state
+                ziti_set_impossible_to_authenticate(ztx);
```
andrewpmartinez, line 1246: A new API session state transition that means we are giving up on this identity.

library/ziti.c
```diff
@@ -1260,6 +1274,7 @@ static void api_session_cb(ziti_api_session *session, const ziti_error *err, voi
             }
         } else {
             ZTX_LOG(DEBUG, "unhandled error, setting api_session_timer to 5s");
+            ziti_set_unauthenticated(ztx);
```
andrewpmartinez, line 1277: Default error transition

library/ziti_ctrl.c
```diff
@@ -33,6 +33,9 @@ const char *const PC_PROCESS_MULTI_TYPE = "PROCESS_MULTI";
 const char *const PC_MAC_TYPE = "MAC";
 const char *const PC_ENDPOINT_STATE_TYPE = "ENDPOINT_STATE";
 
+const char *const ERROR_CODE_UNAUTHORIZED = "UNAUTHORIZED";
+const char *const ERROR_MSG_NO_API_SESSION_TOKEN = "no api session token set for ziti_controller";
```
andrewpmartinez, line 37: Wasn't sure what kind of `ziti_error` to raise. These are used throughout the defensive API session token checks in all requests that require authentication.

inc_internal/ziti_ctrl.h
```diff
@@ -38,14 +38,15 @@ extern const char* const PC_ENDPOINT_STATE_TYPE;
...
-    char *session;
-
+    char *api_session_token;
 } ziti_controller;
 
 int ziti_ctrl_init(uv_loop_t *loop, ziti_controller *ctlr, const char *url, tls_context *tls);
 
 int ziti_ctrl_close(ziti_controller *ctrl);
 
+void ctrl_clear_api_session(ziti_controller *ctrl);
```
ekoby, line 48: make it ziti_ctrl_xxxxx

library/ziti_ctrl.c
```diff
@@ -366,6 +374,16 @@ void ziti_ctrl_login(
 }
 
 void ziti_ctrl_current_identity(ziti_controller *ctrl, void(*cb)(ziti_identity_data *, const ziti_error *, void *), void *ctx) {
+    if(ctrl->api_session_token == NULL) {
```
ekoby, line 377: refactor into a function, something like `bool verify_api_session(cb, ctx)`

if no session it would call `cb` with error return `false`otherwise return `true`

all these could be changed to 
```if (!verify_session(cb, ctx)) return;```

andrewpmartinez: fixed.

inc_internal/ziti_ctrl.h
andrewpmartinez: ah yes good call

ekoby (approved):

PR #315, opened 2021-09-30
Posture Timer Service Events & Forced Service Update Refresh

andrewpmartinez:
adds logic to detect timeout and timeoutRemaining changes on services, forces service refresh post MFA TOTP/posture submission

- timeout changes are detected during normal service refresh loop, compared with previous known value, triggers service changed events
- timeoutRemaning is checked during posture response submissions, the response carries a list of affected services (used for API tests but also works here), services are marked to be forced into the changed status in the subsequent service event
- after posture response and TOTP submissions, a service refresh check is immediately triggered

### DIFF
--- a/programs/ziti-prox-c/proxy.c
+++ b/programs/ziti-prox-c/proxy.c
@@ -534,8 +534,6 @@ const char *my_configs[] = {
         "all", NULL
 };
 
-uv_loop_t *global_loop;
-
 struct mfa_work {
     uv_work_t w;
     ziti_context ztx;
--- a/programs/ziti-prox-c/proxy.c
+++ b/programs/ziti-prox-c/proxy.c
@@ -105,14 +105,6 @@ static void shutdown_timer_cb(uv_timer_t *t) {
     uv_print_active_handles(l, stderr);
 }
 
-void debug_timer(uv_timer_t *t){
-    ziti_context ztx = t->data;
-
-    ziti_endpoint_state_change(ztx, true, false);
-
-    FREE(t);
-}
-
 static void process_stop(uv_loop_t *loop, struct proxy_app_ctx *app_ctx) {
     PREPF(uv, uv_strerror);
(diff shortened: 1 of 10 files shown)


tests/test_ziti_model.cpp
```diff
@@ -32,6 +32,29 @@ limitations under the License.
 
 using Catch::Matchers::Equals;
 
+TEST_CASE("posture response response", "[model]") {
```
dovholuknf, line 35: could be worth testing malformed json to ensure no crashy crash happens... also might be worthwhile to return empty values etc in there for the same reasons. or not ya know, just an idea ;)

dovholuknf: might be worth testing the values for boundary / crazy values too... timeoutRemaining of max int etc. again - just some thoughts

programs/ziti-prox-c/proxy.c
```diff
@@ -85,6 +85,8 @@ struct client {
     LIST_ENTRY(client) next;
 };
 
+uv_loop_t *global_loop;
```
dovholuknf, line 88: there some magic referencing this? is it used elsewhere? i don't see it anywhere else in the pr

dovholuknf (approved):

tests/test_ziti_model.cpp
andrewpmartinez: That is outside the scope of what I added does. That would be the responsibility of the model framework and jsmn.

programs/ziti-prox-c/proxy.c
andrewpmartinez: This is a test program.

andrewpmartinez: Yea I did a bad job of reverting. I moved the declaration up in the file for some debug code. I removed the duplicate.

PR #319, opened 2021-10-08
map enhancements

ekoby:
- support arbitrary keys
- convenience functions for integer/long keys
- avoid memory allocations for short(length <= pointer size) keys

### DIFF
--- a/library/model_support.c
+++ b/library/model_support.c
@@ -981,13 +981,16 @@ int json_enum(const void *ptr, void *bufp, int indent, int flags, const void *en
 
 
 struct model_map_entry {
-    char *key;
+    void *key;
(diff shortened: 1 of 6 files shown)


library/model_support.c
```diff
@@ -1148,7 +1205,9 @@ model_map_iter model_map_it_remove(model_map_iter it) {
         e->_impl->size--;
         LIST_REMOVE(e, _next);
         LIST_REMOVE(e, _tnext);
-        free(e->key);
+        if (e->key_len > sizeof(e->key)) {
+            free(e->key);
```
dovholuknf, line 1209: why wouldn't you want to free this no matter what like before?

PR #327, opened 2021-12-06
make write_buf public for SDK consumers

ekoby:
add printf style buffer write

### DIFF
--- a/includes/ziti/ziti_buffer.h
+++ b/includes/ziti/ziti_buffer.h
@@ -35,17 +35,23 @@ typedef intptr_t ssize_t;
 extern "C" {
 #endif
 
-typedef struct write_buf_s write_buf_t;
+typedef struct string_buf_s string_buf_t;
 
-ZITI_FUNC write_buf_t* new_write_buf();
-ZITI_FUNC write_buf_t* new_fixed_write_buf(char *outbuf, size_t max);
-ZITI_FUNC void delete_write_buf(write_buf_t *wb);
+ZITI_FUNC string_buf_t *new_write_buf();
 
-ZITI_FUNC int write_buf_append(write_buf_t *wb, const char *str);
-ZITI_FUNC int write_buf_append_byte(write_buf_t *wb, char c);
-ZITI_FUNC int write_buf_fmt(write_buf_t *wb, FORMAT_STRING(const char *fmt), ...) ziti_printf_args(2,3);
-ZITI_FUNC size_t write_buf_size(write_buf_t *wb);
-ZITI_FUNC char *write_buf_to_string(write_buf_t *wb, size_t *outlen);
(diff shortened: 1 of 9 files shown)


includes/ziti/ziti_buffer.h
```diff
@@ -0,0 +1,54 @@
...
+#endif
+
+typedef struct write_buf_s write_buf_t;
+
+ZITI_FUNC write_buf_t* new_write_buf();
+ZITI_FUNC write_buf_t* new_fixed_write_buf(char *outbuf, size_t max);
+ZITI_FUNC void delete_write_buf(write_buf_t *wb);
+
+ZITI_FUNC int write_buf_append(write_buf_t *wb, const char *str);
+ZITI_FUNC int write_buf_append_byte(write_buf_t *wb, char c);
```
dovholuknf: since it's public, why not keep this as just `write_buff_append` vs `write_buff_append_string`? also - probably need some kinds of doc. is *str NULL terminated, what is the behavior on overflow stuffs like that

includes/ziti/ziti_buffer.h
```diff
@@ -0,0 +1,54 @@
...
+#endif
+#endif
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct write_buf_s write_buf_t;
+
+ZITI_FUNC write_buf_t* new_write_buf();
```
dovholuknf: doc default size

ekoby: what default size?

includes/ziti/ziti_buffer.h
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * string_buf allocates memory internally as needed unless it was created with [new_fixed_string_buf()]
+ */
+typedef struct string_buf_s string_buf_t;
+
+/**
+ * Create new string buffer.
+ * @return new buffer instance
+ */
+ZITI_FUNC string_buf_t *new_string_buf();
```
dovholuknf, line 49: no tests for: new_string_buf, new_fixed_string_buf, delete_string_buf, string_buf_append_byte

includes/ziti/ziti_buffer.h
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * Deallocate all memory associated with the given string buffer.
+ *
+ * The passed in pointer is not valid after the function invocation.
+ * @param wb
+ */
+ZITI_FUNC void delete_string_buf(string_buf_t *wb);
+
+/**
+ * Append `\0` terminated string to the buffer.
```
dovholuknf, line 68: as an API reader, I wonder, what would happen if i were to pass a string which is not null terminated. would you append it for me or will it end up allowing me to to read past the buffer after i append a non-terminated string?

dovholuknf (approved):
lgtm but adding some of those tests would be swell

PR #328, opened 2021-12-07
better handling of uninitialized/default Enums

ekoby:
- make 0 unset/default value, i.e. Enum_Unknown
- defined Enum values will start at 1
- serialize unset/0 Enum value to JSON as `null`

### DIFF
--- a/library/model_support.c
+++ b/library/model_support.c
@@ -816,6 +816,10 @@ static int _cmp_map(model_map *lh, model_map *rh) {
     return rc;
 }
 
+static int null_to_json(string_buf_t *buf, int indent, int flags) {
+    return string_buf_append(buf, "null");
+}
+
 static int bool_to_json(bool *v, string_buf_t *buf, int indent, int flags) {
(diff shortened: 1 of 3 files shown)


library/model_support.c
```diff
@@ -816,6 +816,10 @@ static int _cmp_map(model_map *lh, model_map *rh) {
     return rc;
 }
 
+static int null_to_json(string_buf_t *buf, int indent, int flags) {
+    return string_buf_append(buf, "null");
```
dovholuknf, line 820: why use "null" and not an empty json string? "{}" isn't an empty json block more like 'null'?

dovholuknf: oh - i see what you did from the tests now. got it

dovholuknf (approved):

library/model_support.c
ekoby: empty object `{}` or empty string are not the same is `null`. `null` is explicit _not set_ value

PR #329, opened 2021-12-09
enroll using jwt content

mary-dcouto:
ziti_enroll function currently accepts jwt file name and read the content from the file. We have added another function which accepts jwt content instead of jwt file name. So  when we call load_jwt_content function, it expects the jwt raw content in the enroll_cfg_s object

### DIFF
--- a/library/jwt.c
+++ b/library/jwt.c
@@ -24,6 +24,57 @@ limitations under the License.
 
 const char* ZITI_SDK_JWT_FILE = "ZITI_SDK_JWT_FILE";
 
+int parse_jwt_content(const char *filename, struct enroll_cfg_s *ecfg, ziti_enrollment_jwt_header **zejh, ziti_enrollment_jwt **zej) {
+
+    const char *dot1 = strchr(ecfg->raw_jwt, '.');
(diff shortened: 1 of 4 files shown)


library/jwt.c
```diff
@@ -24,50 +24,24 @@ limitations under the License.
...
-    if (rc < 0) {
-        ZITI_LOG(WARN, "failed to read JWT file: %d(%s)", errno, strerror(errno));
-        fclose(jwt_input);
-        return ZITI_JWT_INVALID;
-    }
-    jwt_file_len = rc;
-    fclose(jwt_input);
-
-    ZITI_LOG(DEBUG, "jwt file content is: \n%.*s", jwt_file_len, ecfg->raw_jwt);
+int parse_jwt_content(const char *filename, struct enroll_cfg_s *ecfg, ziti_enrollment_jwt_header **zejh, ziti_enrollment_jwt **zej) {
```
ekoby, line 27: don't need filename here, it is only used to log an error

mary-dcouto: removed filename from this function

ekoby (approved):

PR #330, opened 2021-12-10
change parse_XXX() functions to return number of parsed bytes

ekoby:


library/model_support.c
```diff
@@ -142,44 +142,56 @@ int model_cmp(const void *lh, const void *rh, type_meta *meta) {
...
+            goto done;
         }
-        arr[i] = el;
         tok += rc;
     }
     *arrp = arr;
+    done:
+    if (result < 0) {
+        if (arr != NULL) {
+            for (int i = 0; i < children; i++) {
```
dovholuknf, line 174: this code isn't a already somewhere?

dovholuknf (approved):

PR #344, opened 2022-01-31
Created zping example using csdk.

r-caamano:


ekoby:
this probably does not belong here, maybe move into the kitchen

PR #352, opened 2022-02-09
alters service update logic to include updatedAt for posture queries

andrewpmartinez:
Posture queries can be altered by state changes that are reflected in
the updatdAt field on them. This includes timeout changes/reductions for
MFA prompt on wake/unlock as well as for reconfiguration of underlying
posture checks.

This changes the c-sdk to provide service update events for those
changes.


includes/ziti/ziti_model.h
```diff
@@ -69,7 +69,8 @@ XX(query_type, string, none, queryType, __VA_ARGS__) \
 XX(process, ziti_process, ptr, process, __VA_ARGS__) \
 XX(processes, ziti_process, array, processes, __VA_ARGS__) \
 XX(timeout, int, none, timeout, __VA_ARGS__) \
-XX(timeoutRemaining, int, ptr, timeoutRemaining, __VA_ARGS__)
+XX(timeoutRemaining, int, ptr, timeoutRemaining, __VA_ARGS__) \
+XX(updated_at,string, none, updatedAt, __VA_ARGS__)
```
ekoby, line 73: why not timestamp?

andrewpmartinez: No reason other than we currently don't actually care about parsing the values just whether they are different.

The output from the API is an ISO 8601 string - so it could be parsed if needed later.

PR #357, opened 2022-03-10
implement ziti_conn write queue

ekoby:
Allow to _queue_ write requests before connection is connected

### DIFF
--- a/inc_internal/zt_internal.h
+++ b/inc_internal/zt_internal.h
@@ -118,6 +118,8 @@ struct ziti_write_req_s {
     uv_timer_t *timeout;
 
     void *ctx;
+
+    TAILQ_ENTRY(ziti_write_req_s) _next;
 };
 
 struct ziti_conn {
@@ -136,13 +138,14 @@ struct ziti_conn {
     ziti_close_cb close_cb;
     conn_state state;
     bool fin_sent;
-    bool fin_recv;
+    int fin_recv; // 0 - not received, 1 - received, 2 - called app data cb
     bool close;
     int timeout;
 
     buffer *inbound;
     uv_check_t *flusher;
     uv_async_t *disconnector;
+    TAILQ_HEAD(, ziti_write_req_s) wreqs;
     int write_reqs;
 
     void *data;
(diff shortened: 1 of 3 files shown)


inc_internal/zt_internal.h
```diff
@@ -136,13 +138,14 @@ struct ziti_conn {
     ziti_close_cb close_cb;
     conn_state state;
     bool fin_sent;
-    bool fin_recv;
+    int fin_recv; // 0 - not received, 1 - received, 2 - called app data cb
```
scareything, line 141: any reason not to use an enum?

dovholuknf (approved):

library/connect.c
```diff
@@ -572,21 +576,19 @@ static void ziti_write_timeout(uv_timer_t *t) {
     uv_close((uv_handle_t *) t, free_handle);
 }
 
-static void ziti_write_async(uv_async_t *ar) {
-    struct ziti_write_req_s *req = ar->data;
+static void ziti_write_req(struct ziti_write_req_s *req) {
     struct ziti_conn *conn = req->conn;
 
-    if (conn->state >= Disconnected) {
+    if (conn->state >= Timedout) {
```
scareything, line 582: why not go down to `CloseWrite`?

library/connect.c
```diff
@@ -603,20 +605,6 @@ static void ziti_write_async(uv_async_t *ar) {
             send_message(conn, ContentTypeData, req->buf, req->len, req);
         }
     }
-    uv_close((uv_handle_t *) ar, free_handle);
-}
-
-int ziti_write_req(struct ziti_write_req_s *req) {
-    NEWP(ar, uv_async_t);
-    uv_async_init(req->conn->ziti_ctx->loop, ar, ziti_write_async);
```
scareything, line 611: this code is being removed, but I think `uv_async_init` should have been serialized if being called from another thread, since it modifies the loop?

scareything (approved):

library/connect.c
ekoby: yes, it was a bad pattern

library/connect.c
ekoby: `CloseWrite` can be set while there is still data in the queue. FIN will be sent after all data is sent

PR #361, opened 2022-03-28
fix: use idle handle for ziti_conn.flusher to avoid IO wait delay

ekoby:


library/connect.c
```diff
@@ -709,39 +711,47 @@ static int send_crypto_header(ziti_connection conn) {
...
             ziti_write_req(req);
         }
+        CONN_LOG(TRACE, "flushed %d messages", count);
     }
+
+    return !TAILQ_EMPTY(&conn->wreqs);
 }
 
-static void flush_to_client(ziti_connection conn) {
+static bool flush_to_client(ziti_connection conn) {
```
scareything, line 754: nit. If I'm not mistaken the return value indicates whether or not data is still pending, which is the opposite of how I naturally read the signature - whether or not the flush completed. IMO it would be more clear if the function returned the number of bytes remaining.

scareything (approved):

PR #362, opened 2022-03-28
Ziti Connection bridge

ekoby:
Ziti Connection bridge functions (see ziti.h) are utilities to simplify handling of application streams and IO file descriptors

### DIFF
--- a/includes/ziti/ziti.h
+++ b/includes/ziti/ziti.h
@@ -311,7 +311,7 @@ typedef struct ziti_listen_opts_s {
  * @return indicate how much data was consumed
  * @see ziti_dial(), ziti_accept(), ZITI_ERRORS
  */
-typedef ssize_t (*ziti_data_cb)(ziti_connection conn, uint8_t *data, ssize_t length);
+typedef ssize_t (*ziti_data_cb)(ziti_connection conn, const uint8_t *data, ssize_t length);
 
 /**
  * @brief Connection callback.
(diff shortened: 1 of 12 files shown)


includes/ziti/ziti.h
```diff
@@ -800,7 +811,36 @@ extern int ziti_close_write(ziti_connection conn);
...
+ *
+ * @param conn
+ * @param input
+ * @param output
+ * @param close_cb
+ * @param ctx
+ * @return
+ */
+ZITI_FUNC
+extern int ziti_conn_bridge_fds(ziti_connection conn, uv_os_fd_t input, uv_os_fd_t output, void (*close_cb)(void *ctx), void *ctx);
```
smilindave26, line 843: why a different close_cb than on the stream version (and why doesn't the stream bridge include a void *ctx)?

ekoby: with the stream version we don't make any assumptions on how application want to close it -- so it is passed into the close callback. any custom data can be attached to that stream already `stream->data`

library/conn_bridge.c
```diff
@@ -0,0 +1,190 @@
...
+    br->data = br;
+    br->fdbr = fdbr;
+
+    uv_handle_set_data((uv_handle_t *) br->input, br);
+    ziti_conn_set_data(conn, br);
+
+    ziti_conn_set_data_cb(conn, on_ziti_data);
+    uv_read_start(br->input, bridge_alloc, on_input);
+
+    return ZITI_WTF;
```
smilindave26, line 114: really return wtf?  expect to run forever (nobody calls uv_read_stop?)?  should return based on the return value from uv_read_start?

ekoby: Umm... leftover TODO code. good catch

smilindave26 (approved):

PR #364, opened 2022-04-04
don't set partially authenticated if there are no auth queries

### LINKED ISSUE #363: Authenticated sessions needlessly set to partially authenticated
right now when you launch a sample program, the programs check the `ctrl_status` and this status is evented as "partially authenticated" at the wrong time.  Update the csdk to not emit that event if not accurate

dovholuknf:
closes #363

### DIFF
--- a/library/ziti.c
+++ b/library/ziti.c
@@ -1337,6 +1337,8 @@ static void api_session_cb(ziti_api_session *session, const ziti_error *err, voi
 
     if (session) {
 
+        ZTX_LOG(DEBUG, "%s successfully => api_session[%s]", ztx->api_session ? "refreshed" : "logged in", session->id);
+
         ziti_set_api_session(ztx, session);
 
         ziti_auth_query_init(ztx);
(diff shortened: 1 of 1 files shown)


andrewpmartinez (approved):

library/ziti.c
```diff
@@ -1336,10 +1337,6 @@ static void api_session_cb(ziti_api_session *session, const ziti_error *err, voi
 
     if (session) {
 
-        update_ctrl_status(ztx, ZITI_PARTIALLY_AUTHENTICATED, NULL);
-
-        ZTX_LOG(DEBUG, "%s successfully => api_session[%s]", ztx->api_session ? "refreshed" : "logged in", session->id);
-
         ziti_set_api_session(ztx, session);
```
dovholuknf, line 1340: `ziti_set_api_session` was always called and has the appropriate test for auth queries inside it. just moved `update_ctrl_status` into there. I'll put that debug message back though. didn't really want to delete that

PR #365, opened 2022-04-05
Ziti socket

ekoby:


library/ziti-socket.c
```diff
@@ -0,0 +1,404 @@
...
+    err = WSAStartup(ver, &data);
+
+    ziti_sock_server = socket(AF_UNIX, SOCK_STREAM, 0);
+    if (ziti_sock_server == INVALID_SOCKET) {
+        fprintf(stderr, "invalid socket: %d", WSAGetLastError());
+    }
+
+    ziti_sock_name.sun_family = AF_UNIX;
+    char temp[sizeof(ziti_sock_name.sun_path)];
+    GetTempPath(sizeof(temp), temp);
```
scareything, line 376: this may return a path that is writable by users other than the caller, and the ownership/permissions of this file determine the permissions of the socket (from [AF_UNIX comes to Windows](https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/#security)). Maybe `%LOCALAPPDATA%` or explicitly creating a directory with the desired permission would be better?

library/ziti-socket.c
```diff
@@ -0,0 +1,404 @@
...
+    snprintf(ziti_sock_name.sun_path, sizeof(ziti_sock_name.sun_path), "%sziti-socket.%d", temp, uv_os_getpid());
```
scareything, line 377: only one ziti socket per process?

includes/ziti/socket.h
```diff
@@ -0,0 +1,56 @@
...
+
+#if _WIN32
+#include <WinSock2.h>
+typedef SOCKET ziti_socket_t;
+#else
+typedef int ziti_socket_t;
+#endif
+
+ZITI_FUNC
+void Ziti_lib_init(void);
```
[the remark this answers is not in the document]
ekoby: capital Z to differentiate from other/lower level API -- this API does not mix with the  other one

library/ziti-socket.c
ekoby: one server listen socket -- unlimited connected client/server sockets

library/ziti-socket.c
ekoby: `GetTempPath()` returns `%LOCALAPPDATA%\Temp` which is not accessible by other users (except admin)

scareything: `GetTempPath` returns `%LOCALAPPDATA%` at least in some cases. [The doc](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppatha) says it may return other values:

> The GetTempPath function checks for the existence of environment variables in the following order and uses the first path found:
> The path specified by the TMP environment variable.
> The path specified by the TEMP environment variable.
> The path specified by the USERPROFILE environment variable.
> The Windows directory.

library/ziti-socket.c
```diff
@@ -0,0 +1,402 @@
...
+        complete_future(f, wrap->ztx);
+    } else {
+        LIST_INSERT_HEAD(&wrap->futures, f, _next);
+    }
+}
+
+ziti_context Ziti_load_context(const char *identity) {
+    future_t *f = schedule_on_loop(load_ziti_ctx, identity, true);
+    int err = await_future(f);
+    errno = err;
```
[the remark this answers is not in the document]
scareything: fwiw I think it's really confusing to use `errno` to express ziti error constants.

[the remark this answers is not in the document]
ekoby: replaced errno with Ziti_last_error()

scareything (approved):

PR #370, opened 2022-04-12
replace deprecated upload action

ekoby:


andrewpmartinez (approved):

.github/workflows/build-release.yml
```diff
@@ -41,20 +41,9 @@ jobs:
...
+        uses: softprops/action-gh-release@v1
+        if: startsWith(github.ref, 'refs/tags/')
         with:
-          upload_url: ${{ steps.get_release.outputs.upload_url }}
-          asset_path: ziti-prox-c-${{ runner.os }}.zip
-          asset_name: ziti-prox-c-${{ runner.os }}.zip
-          asset_content_type: application/octet-stream
+          files: |
+            ziti-prox-c-${{ runner.os }}.zip
+            ${{runner.workspace}}/build/package/ziti-sdk*
```
qrkourier, line 49: Consider also adding the `with` param `fail_on_unmatched_files: true` so that cases where _any_ of the expected files globs are not present.

PR #376, opened 2022-04-29
Ziti socket intercepts

ekoby:
implement `Ziti_connect_addr()` to connect via intercepted address


includes/ziti/socket.h
```diff
@@ -63,7 +63,7 @@ ziti_context Ziti_load_context(const char *identity);
  * @return native socket handle
  */
 ZITI_FUNC
-ziti_socket_t Ziti_socket();
+ziti_socket_t Ziti_socket(int type);
```
dovholuknf, line 66: would you add info for what 'type' means? is it TCP/UDP? or raw or something else?

library/ziti-socket.c
```diff
@@ -187,16 +189,66 @@ static void on_ctx_event(ziti_context ztx, const ziti_event_t *ev) {
                 free(wrap);
             }
         }
+    } else if (ev->type == ZitiServiceEvent) {
+
+
+        for (int i = 0; ev->event.service.removed && ev->event.service.removed[i] != NULL; i++) {
+            ziti_intercept_cfg_v1 *intercept = model_map_remove(&wrap->intercepts, ev->event.service.removed[i]->name);
+            free_ziti_intercept_cfg_v1(intercept);
+            FREE(intercept);
```
dovholuknf, line 198: it's guaranteed that an event will never have removed/changed/added entries, right? otherwise this will double FREE(intercept). Just making sure that's a guarantee...

dovholuknf: ugh. tired... i see it is scoped in the loop... my bad

includes/ziti/socket.h
ekoby: updated

library/ziti-socket.c
```diff
@@ -280,38 +338,135 @@ static void on_bridge_close(void *ctx) {
...
+                                  "{\"dst_protocol\": \"%s\", \"dst_hostname\": \"%s\", \"dst_port\": \"%u\"}",
+                                  proto_str, req->host, req->port);
+            ziti_dial_opts opts = {
+                    .app_data = app_data,
+                    .app_data_sz = len,
+            };
+            ZITI_LOG(INFO, "zs[%p]->f[%p]", zs, zs->f);
+            ziti_dial_with_options(zs->conn, req->service, &opts, on_ziti_connect, NULL);
+        } else {
+            fail_future(f, -ECONNREFUSED);
```
dovholuknf, line 438: i think you should log this too. this seems like a strange situation to be in

ekoby: it is not that strange: the address you're connecting to is not in any service intercepts

library/ziti.c
```diff
@@ -680,8 +678,10 @@ const char *ziti_conn_source_identity(ziti_connection conn) {
 
 
 void ziti_send_event(ziti_context ztx, const ziti_event_t *e) {
-    if ((ztx->opts->events & e->type) && ztx->opts->event_cb) {
-        ztx->opts->event_cb(ztx, e);
+    if (ztx->enabled || e->type == ZitiContextEvent) {
```
dovholuknf, line 681: only ZitiContextEvent?

dovholuknf (approved):

library/ziti.c
ekoby: once ztx is disabled (via ziti_shutdown()) app would not expect other events

PR #382, opened 2022-05-10
check session ref for validity before using it to connect

ekoby:


library/ziti.c
```diff
@@ -1427,6 +1427,15 @@ static void version_cb(ziti_version *v, const ziti_error *err, void *ctx) {
     }
 }
 
+bool ziti_is_session_valid(ziti_context ztx, ziti_net_session *session, const char *service_id, ziti_session_type type) {
+    if (session == NULL) return false;
+
+    if (type == ziti_session_types.Bind) return true;
```
andrewpmartinez, line 1433: Why return true if we can bind?

Just reading the name I would expect this to see if we have a session that is correct and active.

ekoby: `Bind` sessions are treated differently from `Dial` ones. They cannot be shared between ziti connections and are not cached by ziti_context. The actual object is owned by the corresponding connection and should always be valid for the purposes of this function.

PR #383, opened 2022-05-10
add architecture to output package name

ekoby:
add crosscompile for linux_arm
add crosscompile for macOS_arm

install crossbuild tools

update release workflow

add arch to ziti-prox-c zip file

use 7zip is available


.github/workflows/build-release.yml
```diff
@@ -6,44 +6,58 @@ on:
...
-        run: 7z a -tzip ziti-prox-c-${{ runner.os }}.zip ${{runner.workspace}}/build/programs/ziti-prox-c/Debug/ziti-prox-c.exe
-      - name: Get release
-        id: get_release
-        uses: bruceadams/get-release@v1.2.2
-        env:
-          GITHUB_TOKEN: ${{ github.token }}
+        run: cmake --build ${{runner.workspace}}/build --target ziti-prox-c-pack --target package
+
+#      - name: zip artifacts
+#        if: runner.os != 'Windows'
```
scareything, line 47: Could we use `cmake -E tar --format=zip` on all operating systems?

ekoby: not necessarily, support for `--format=zip` is not guaranteed.
In practice it is only used by our CI

scareything (approved):

PR #393, opened 2022-05-24
Ziti bridge throttling

ekoby:
make sure we don't overwhelm the loop if bridge receives too much data
avoid crypto failures (due to fragmentation) on older networks with large message sizes

### DIFF
--- a/library/conn_bridge.c
+++ b/library/conn_bridge.c
@@ -52,7 +52,7 @@ extern int ziti_conn_bridge(ziti_connection conn, uv_stream_t *stream, uv_close_
     br->output = stream;
     br->close_cb = on_close;
     br->data = uv_handle_get_data((const uv_handle_t *) stream);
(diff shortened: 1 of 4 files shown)


library/conn_bridge.c
```diff
@@ -48,6 +52,7 @@ extern int ziti_conn_bridge(ziti_connection conn, uv_stream_t *stream, uv_close_
     br->output = stream;
     br->close_cb = on_close;
     br->data = uv_handle_get_data((const uv_handle_t *) stream);
+    br->input_pool = pool_new(BRIDGE_MSG_SIZE, 16, NULL);
```
smilindave26, line 55: BRIDGE_POOL_SIZE?

ekoby: yep

smilindave26 (approved):

PR #418, opened 2022-06-23
keep ztx timers during disable/enable cycle

ekoby:


inc_internal/posture.h
```diff
@@ -26,7 +26,6 @@ extern "C" {
 
 struct posture_checks {
     uv_timer_t *timer;
-    double interval;
```
ekoby, line 29: this was set but not used anywhere

smilindave26 (approved):

PR #422, opened 2022-06-30
allow objects returned to pool after pool_destroy()

ekoby:


library/pool.c
```diff
@@ -31,41 +31,46 @@ struct pool_obj_s {
...
+
+    while (!LIST_EMPTY(&pool->pool)) {
+        struct pool_obj_s *m = LIST_FIRST(&pool->pool);
         LIST_REMOVE(m, _next);
         free(m);
     }
 
-    free(p);
+    if (pool->out == 0)
+        free(pool);
```
dovholuknf, line 59: why the 'if' here? something else must eventually end up free'ing the pool?

library/pool.c
```diff
@@ -31,41 +32,47 @@ struct pool_obj_s {
...
 }
 
 bool pool_has_available(pool_t *pool) {
-    return !LIST_EMPTY(&pool->pool) || pool->count > 0;
+    assert(!pool->is_closed);
```
smilindave26, line 64: that's pretty violent.  better to do WTF log?

library/pool.c
```diff
@@ -31,41 +32,47 @@ struct pool_obj_s {
...
+    return !LIST_EMPTY(&pool->pool) || pool->capacity > pool->out;
 }
 
 void *pool_alloc_obj(pool_t *pool) {
+    assert(!pool->is_closed);
```
smilindave26, line 69: more violence!

library/pool.c
ekoby: this flags a programming error, i.e. should never happen

PR #428, opened 2022-07-13
add README.me, implement enrollment with zitify

ekoby:


zitify/zitify
```diff
@@ -19,4 +19,65 @@
...
+usage() {
+    echo "zitify <cmd> [args]"
+    echo "       Run zitified <cmd>"
+    echo ""
+    echo "zitify enroll -j <enrollment token> -i <identity file> [ -k|--key <private key> [ -c|--cert certificate ] ]"
+    echo "       Enroll identity"
+}
+
+function enroll() {
+  python - $* <<EOF
```
smilindave26, line 31: I don't think we support macOS yet, but will we see systems that need to use `python3`?

zitify/README.md
```diff
@@ -0,0 +1,26 @@
...
+## Try the Zitify
+
+Download the latest release and extract into a directory on your `$PATH`
+
+Acquire an enrollment token from [ZEDS](https://zeds.openziti.org)
+
+Enroll (assume your file is `my_id.jwt`):
+
+```console
+$ zitify enroll -j my_id.jwt -i my_id.json
```
smilindave26, line 18: if we require python (and any modules that need to be installed), we should call that out

smilindave26 (approved):

PR #431, opened 2022-07-15
fix process posture check

### LINKED ISSUE #430: posture checks not evaluating all processes on Windows
Some processes are not allowed to be inspected, breaking certain posture checks.

dovholuknf:
closes #430 

* use PROCESS_QUERY_LIMITED_INFORMATION instead of PROCESS_QUERY_INFORMATION
* remove unused `ziti_enable_uv_mbed_logger`
* change posture checks to fire 1s after startup instead of 20
* fix some %d -> %lu warnings
* Add a WARN message when the process cannot be enumerated
* change `%FT%T` to `%Y-%m-%dT%H:%M:%S` - this was constantly throwing warnings when observed in gdb and apparently doesn't work on windows with mingw

### DIFF
--- a/library/posture.c
+++ b/library/posture.c
@@ -930,7 +930,9 @@ static bool check_running(uv_loop_t *loop, const char *path) {
 
         HANDLE ph = OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION, FALSE, pe32.th32ProcessID); //OpenProcess(PROCESS_QUERY_INFORMATION, FALSE, pe32.th32ProcessID);
         if (ph == NULL) {
-            ZITI_LOG(WARN, "process %s is running, however not able to open handle. GetLastError(): %lu", pe32.szExeFile, GetLastError());
+            if (pe32.th32ProcessID > 0) {
+                ZITI_LOG(DEBUG, "process %s is running, however not able to open handle. GetLastError(): %lu", pe32.szExeFile, GetLastError());
+            }
             continue;
         }
         fullPathSize = sizeof(fullPath);
--- a/library/posture.c
+++ b/library/posture.c
@@ -911,7 +911,7 @@ static bool check_running(uv_loop_t *loop, const char *path) {
 
     // Set the size of the structure before using it.
     PROCESSENTRY32 pe32;
-    pe32.dwSize = sizeof( PROCESSENTRY32 );
+    pe32.dwSize = sizeof(pe32);
     // Retrieve information about the first process, and exit if unsuccessful
     if( !Process32First( sh, &pe32 ) )
     {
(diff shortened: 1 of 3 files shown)


library/posture.c
```diff
@@ -129,7 +129,7 @@ void ziti_posture_init(ziti_context ztx, long interval_secs) {
     }
 
     if (!uv_is_active((uv_handle_t *) ztx->posture_checks->timer)) {
-        uv_timer_start(ztx->posture_checks->timer, ziti_pr_ticker_cb, MILLIS(interval_secs), MILLIS(interval_secs));
+        uv_timer_start(ztx->posture_checks->timer, ziti_pr_ticker_cb, MILLIS(1)/*fire on startup*/, MILLIS(interval_secs));
```
scareything, line 132: You could use `0` here, right?

http://docs.libuv.org/en/v1.x/timer.html#c.uv_timer_start

library/posture.c
```diff
@@ -905,29 +905,46 @@ static bool check_running(uv_loop_t *loop, const char *path) {
...
+        ZITI_LOG(ERROR, "failed to get process list: %lu", GetLastError());
         return result;
     }
-    PROCESSENTRY32W entry = {
-            .dwSize = sizeof(PROCESSENTRY32W)
-    };
+
+    // Set the size of the structure before using it.
+    PROCESSENTRY32 pe32;
+    pe32.dwSize = sizeof( PROCESSENTRY32 );
```
scareything, line 914: nit/style: `sizeof(pe32)` is always correct, even if the variable's type is changed in the future.

library/posture.c
```diff
@@ -905,29 +905,46 @@ static bool check_running(uv_loop_t *loop, const char *path) {
...
 
-    for (BOOL ret = Process32FirstW(sh, &entry); ret; ret = Process32NextW(sh, &entry)) {
-        HANDLE ph = OpenProcess(PROCESS_QUERY_INFORMATION, FALSE, entry.th32ProcessID);
+    // Now walk the snapshot of processes, and display information about each process in turn
+    ZITI_LOG(VERBOSE, "checking to see if process is running: %s", path);
+    do
+    {
+        ZITI_LOG(VERBOSE, "process is running: %s", pe32.szExeFile);
+
+        HANDLE ph = OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION, FALSE, pe32.th32ProcessID); //OpenProcess(PROCESS_QUERY_INFORMATION, FALSE, pe32.th32ProcessID);
```
scareything, line 931: probably safe to remove the comment on this line?

library/utils.c
```diff
@@ -371,7 +368,7 @@ void ziti_fmt_time(char* time_str, size_t time_str_sz, uv_timeval64_t* tv) {
         strncpy(time_str, "null tv", time_str_sz);
     } else {
         struct tm* start_tm = gmtime(&tv->tv_sec);
-        strftime(time_str, time_str_sz, "%FT%T", start_tm);
+        strftime(time_str, time_str_sz, "%Y-%m-%dT%H:%M:%S", start_tm);
```
scareything, line 371: yay!

scareything (approved):

library/posture.c
dovholuknf: fixed. thx

library/posture.c
dovholuknf: fixed. thx. i straight up "forklifted" the code, but it makes a lot of sense :)

PR #432, opened 2022-07-18
same commit but should be signed. see https://github.com/openziti/zit…

### LINKED ISSUE #430: posture checks not evaluating all processes on Windows

dovholuknf:
closes #430 previous PR had an unsigned commit somehow :/ 

* use PROCESS_QUERY_LIMITED_INFORMATION instead of PROCESS_QUERY_INFORMATION
* remove unused `ziti_enable_uv_mbed_logger`
* change posture checks to fire 1s after startup instead of 20
* fix some %d -> %lu warnings
* Add a WARN message when the process cannot be enumerated
* change `%FT%T` to `%Y-%m-%dT%H:%M:%S` - this was constantly throwing warnings when observed in gdb and apparently doesn't work on windows with mingw

### DIFF
--- a/includes/ziti/ziti_log.h
+++ b/includes/ziti/ziti_log.h
@@ -101,11 +101,6 @@ ZITI_FUNC extern const char* ziti_log_level_label();
  */
 ZITI_FUNC void uv_mbed_logger(int level, const char *file, unsigned int line, const char *msg);
 
-// don't use this function. it's a temporary function to support logging on windows.it will be
-// removed in a future release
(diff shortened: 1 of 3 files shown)


scareything (approved):
For the record I still think it’s silly to start a uv timer with a 1ms delay

dovholuknf:
while I don't disagree, it's actually 1000ms, and I didn't want to rock this boat all that much. I reduced the delay from 20s to 1s

PR #433, opened 2022-07-18
allow to specify targeted(module, or file) log levels

ekoby:
example
```
export ZITI_LOG="2;ziti-sdk=3;ziti_ctrl.c=6"
```
set all logging to WARN(2), sdk logging to INFO(3), except ziti controller which is set to TRACE(6)

you can also enable uv-mbed logging by specifying `uv-mbed` with desired level: `uv-mbed=4`

### DIFF
--- a/library/utils.c
+++ b/library/utils.c
@@ -71,6 +71,8 @@ static const char *const level_labels[] = {
         DEBUG_LEVELS(LEVEL_LBL)
 };
 
+static const char *basename(const char *path);
+
 const char *ziti_get_build_version(int verbose) {
     if (verbose) {
         return "\n\tVersion:\t" to_str(ZITI_VERSION)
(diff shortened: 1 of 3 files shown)


library/utils.c
```diff
@@ -141,30 +140,43 @@ void ziti_log_init(uv_loop_t *loop, int level, log_writer log_func) {
...
+        int len = snprintf(msg, sizeof(msg), "set log level: %s=%d", marker ? marker : "root", level);
         logger(INFO, "ziti_log_set_level", msg, len);
     }
 }
 
-int ziti_log_level() {
+int ziti_log_level(const char *module, const char *file) {
+    int level;
+
+    file = basename(file);
```
smilindave26, line 171: basename only returns NULL if there's an error (If `file` is passed in as NULL, "." is returned).  Is this what you're going for?

ekoby: > basename only returns NULL if there's an error (If `file` is passed in as NULL, "." is returned). Is this what you're going for?

we don't use standard basename() function

smilindave26 (approved):

PR #434, opened 2022-07-18
fix memory leaks and ztx disabling

ekoby:


library/posture.c
```diff
@@ -814,7 +836,7 @@ static void process_check_work(uv_work_t *w) {
     pcw->is_running = check_running(w->loop, path);
     if (hash_sha512(ztx, w->loop, path, &digest, &digest_len) == 0) {
         hexify(digest, digest_len, 0, &pcw->sha512);
-        ZTX_LOG(VERBOSE, "file(%s) hash = %s", path, pcw->sha512);
+        ZITI_LOG(VERBOSE, "file(%s) hash = %s", path, pcw->sha512);
```
andrewpmartinez, line 839: why change these from ZTX -> ZITI?

ekoby: this executes on the worker thread, so it is possible that ztx pointer became invalid (if ztx was shutdown)

PR #453, opened 2022-09-27
use intercepts in ziti_src

ekoby:
ziti_src can be initialized without a service -- service will be looked up based on intercept configs
also make sure we provide forwarding data in dial options

### DIFF
--- a/library/ziti_src.c
+++ b/library/ziti_src.c
@@ -46,7 +46,7 @@ const char *zl_strerror(uv_link_t *link, int err);
 
 static void zlnf_conn_cb(ziti_connection conn, int status);
 
-static ssize_t zlnf_data_cb(ziti_connection conn, uint8_t *data, ssize_t length);
+static ssize_t zlnf_data_cb(ziti_connection conn, const uint8_t *data, ssize_t length);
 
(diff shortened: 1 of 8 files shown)


library/ziti_src.c
```diff
@@ -74,16 +74,40 @@ int ziti_src_init(uv_loop_t *l, um_src_t *st, const char *svc, ziti_context ztx)
...
 
+    char app_data[1024];
+    size_t app_data_len = snprintf(app_data, sizeof(app_data), "{"
+                                         "\"dst_protocol\":\"tcp\","
+                                         "\"dst_hostname\":\"%s\", "
+                                         "\"dst_port\":\"%s\""
+                                         "}", host, port);
+    ziti_dial_opts opts = {
+            .app_data = app_data,
+            .app_data_sz = app_data_len,
```
scareything, line 98: we should probably check that `snprintf` returned a positive value before getting here.

ekoby: `snprintf` always returns a positive value

scareything: ahh, TIL C99 compliant implementations don't return -1 when the buffer is too small.

smilindave26 (approved):

Issue #416, opened 2022-06-22
crash when ziti identity is invalid


dovholuknf:
during the ip->dns migration many users had "garbage" replaced in the identity file. When there's an invalid URL in the ztAPI, the SDK crashes. It should not crash.

![image](https://user-images.githubusercontent.com/46322585/174941416-a9e442d9-13a4-4f3d-b9e8-0a1fadca17a8.png)

0xc0dedad:
@dovholuknf,
I'd like to handle this issue. Please assign it to me.

dovholuknf:
You got it. Thanks!

PR #465, opened 2022-10-22
Terminate app due to ziti_ctrl_init fail

0xc0dedad:
Fix for https://github.com/openziti/ziti-sdk-c/issues/416

The controller URL is validating in scope of um_http_init. So, the termination procedures for ziti_ctrl_init have added.

### DIFF
--- a/library/ziti.c
+++ b/library/ziti.c
@@ -413,6 +413,9 @@ static void ziti_init_async(ziti_context ztx, void *data) {
     ztx->id = ztx_seq++;
     uv_loop_t *loop = ztx->w_async.loop;
     struct ziti_init_req *init_req = data;
+    ziti_event_t ev = {
+	     .type = ZitiContextEvent,
+    };
 
     uv_timeval64_t start_time;
     uv_gettimeofday(&start_time);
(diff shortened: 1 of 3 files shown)


programs/sample_enroll/sample_enroll.c
```diff
@@ -75,6 +75,12 @@ struct enroll_cert {
 #include <ziti/zitilib.h>
 
 int main(int argc, char **argv) {
+
+    if (argc != 2) {
```
ekoby, line 79: is this from the other PR?

0xc0dedad: Oops, sorry. I accidentally push to this branch also..

0xc0dedad: Redundant commit is removed.

library/ziti.c
```diff
@@ -426,7 +426,12 @@ static void ziti_init_async(ziti_context ztx, void *data) {
     ZTX_LOG(INFO, "using uv_mbed[%s], tls[%s]", uv_mbed_version(), ztx->tlsCtx->api->version ? ztx->tlsCtx->api->version() : "unspecified");
     ZTX_LOG(INFO, "Loading from config[%s] controller[%s]", ztx->opts->config, ztx_controller(ztx));
 
-    ziti_ctrl_init(loop, &ztx->controller, ztx_controller(ztx), ztx->tlsCtx);
+    if (ziti_ctrl_init(loop, &ztx->controller, ztx_controller(ztx), ztx->tlsCtx) != ZITI_OK) {
+	ZITI_LOG(ERROR, "Ziti controller init failed");
+	uv_stop(loop);
```
ekoby, line 431: remove this -- the loop is not controlled by the ziti_context, it is provided by the code using SDK.

ekoby (changes requested):

library/ziti.c
0xc0dedad: I'm using sample_host for testing invalid config file. The app will be stuck in loop and do nothing if uv_stop don't call.
Could you please advise me how to properly propagate the error to control code? Or such behavior is sufficient in this case?

ekoby: without callback application won't receive the error

0xc0dedad: Changed to ziti_send_event

ekoby (approved):
thank you for the contribution

PR #466, opened 2022-10-23
Args prompt for sample_enroll

0xc0dedad:


programs/sample_enroll/sample_enroll.c
```diff
@@ -75,6 +75,12 @@ struct enroll_cert {
 #include <ziti/zitilib.h>
 
 int main(int argc, char **argv) {
+
+    if (argc != 2) {
```
ekoby, line 79: this should be `argc >= 3`

usage is `sample_enroll <JWT file> <ID file> [key_file] [cert_file]`

ekoby (changes requested):
see comments

programs/sample_enroll/sample_enroll.c
0xc0dedad: You probably have meant `if (argc  < 3) -> prompt`?
JWT/ID files are mandatory.
key/cert files are optional.

0xc0dedad: Fixed.

ekoby (approved):

ekoby:
Thank you for your contribution

PR #469, opened 2022-10-25
make sure ziti_listen/ziti_connect fail with appropriate message on missing permission

### LINKED ISSUE #467: ziti_errorstr yeilds programming error after calling zitilib Ziti_bind when identity is not authorized to bind
I called Ziti_bind with an identity which was not authorized to bind. Ziti_bind returns -1 response and ziti_errorstr returns erroNo -34, "WTF: programming error".

## Steps to reproduce:
1. make an identity which is able to dial a service
1. attempt to use the identity to bind a service. call:
    1. Ziti_load_context [succeeds]
    2. Ziti_socket [succeeds]
    3. Ziti_bind [fails]
    4. Ziti_last_error [error -34]

## Expected Behavior:
I would have expected a different error such as "not authorized to bind service" or something similar

ekoby:
[fixes #467]


library/internal_model.c
```diff
@@ -129,6 +130,17 @@ IMPL_ENUM(ziti_session_type, ZITI_SESSION_TYPE_ENUM)
...
 
+bool ziti_service_has_permission(const ziti_service *service, ziti_session_type sessionType) {
+    if (sessionType == ziti_session_types.Dial) {
+        return (service->perm_flags & ZITI_CAN_DIAL) != 0;
+    }
+
+    if (sessionType == ziti_session_types.Bind) {
+        return (service->perm_flags & ZITI_CAN_BIND) != 0;
+    }
+
```
dovholuknf, line 141: maybe a log statement here indicating that this is an unexpected scenario? something like that?

dovholuknf (approved):
i think it'd be good to add the log statement if we hit the unexpected scenario. other than that, lgtm

library/internal_model.c
ekoby: this is just a check, log message is where it is used

PR #487, opened 2023-01-24
allow config string to be the config JSON [fixes #462]

### LINKED ISSUE #462: Ziti_load_context and ziti_init should take a byte array/function - not just a file
Ziti_load_context currently expects to load a file from the drive. That's helpful but it'd be as helpful to take a byte array or a function to return a byte array or something that would allow me to pass something other than a file

ekoby:


library/config.c
```diff
@@ -56,31 +52,30 @@ int load_config_file(const char *filename, ziti_config **cfg) {
...
+int load_config(const char *cfgstr, ziti_config *cfg) {
+    if (!cfgstr) {
+        return ZITI_INVALID_CONFIG;
+    }
+
+    memset(cfg, 0, sizeof(*cfg));
+    int rc = parse_ziti_config(cfg, cfgstr, strlen(cfgstr));
+
+    if (rc < 0) {
+        rc = load_config_file(cfgstr, cfg);
```
scareything, line 73: maybe a DEBUG log here so we can tell that config parsing failed and are falling back to file loading?

scareything (approved):

PR #488, opened 2023-01-27
implement ziti bridge for UDP handles

ekoby:


includes/ziti/ziti.h
```diff
@@ -825,12 +825,21 @@ extern int ziti_write(ziti_connection conn, uint8_t *data, size_t length, ziti_w
...
+
+/**
+ * set idle timeout on bridged connection.
+ * @param conn ziti_connection previously bridged wtih [ziti_conn_bridge] or [ziti_conn_bridge_fds]
+ * @param millis timeout after which bridge will be closed absent any traffic
+ * @return 0 on success, error code on failure
  */
 ZITI_FUNC
-extern int ziti_conn_bridge(ziti_connection conn, uv_stream_t *stream, uv_close_cb on_close);
+extern int ziti_conn_bridge_idle_timeout(ziti_connection conn, unsigned long millis);
```
dovholuknf, line 842: why not use a struct for params instead? also this is changing the api, should come along with a version bump

library/conn_bridge.c
```diff
@@ -32,36 +32,60 @@ struct ziti_bridge_s {
...
 
 static void on_input(uv_stream_t *s, ssize_t len, const uv_buf_t *b);
+static void on_udp_input(uv_udp_t *udp, ssize_t len, const uv_buf_t *b, const struct sockaddr *addr, unsigned int flags);
 
+extern int ziti_conn_bridge(ziti_connection conn, uv_handle_t *handle, uv_close_cb on_close) {
+    if (handle == NULL) return UV_EINVAL;
+
+    if ( !(handle->type == UV_TCP || handle->type == UV_NAMED_PIPE ||
+           handle->type == UV_TTY || handle->type == UV_UDP )) {
+        return UV_EINVAL;
```
dovholuknf, line 59: i feel like this would be a useful place for a log message telling the consumer what/why it's not working. i see it returns an error code, but it might be helpful. also this logic should be reflected in the doc

library/conn_bridge.c
```diff
@@ -124,15 +165,41 @@ extern int ziti_conn_bridge_fds(ziti_connection conn, uv_os_fd_t input, uv_os_fd
...
+int ziti_conn_bridge_idle_timeout(ziti_connection conn, unsigned long millis) {
+    struct ziti_bridge_s *br = ziti_conn_data(conn);
+    if (millis == 0) {
+        br->idle_timeout = 0;
+        if (br->idle_timer) {
+            uv_close((uv_handle_t *) br->idle_timer, (uv_close_cb) free);
+            br->idle_timer = NULL;
+        }
+    } else {
+        br->idle_timeout = millis;
```
dovholuknf, line 192: i feel like this should have a floor too. we shouldn't allow values lower than some threshold, or at very least log a warning when/if that happens. That'll be hard to track down if someone puts a value very low in there

library/conn_bridge.c
```diff
@@ -163,6 +235,16 @@ static void on_output(uv_write_t *wr, int status) {
     free(wr);
 }
 
+static void on_udp_send(uv_udp_send_t *sr, int status) {
+    if (status != 0) {
+        struct ziti_bridge_s *br = sr->handle->data;
+        ZITI_LOG(WARN, "udp_send failed: %d(%s)", status, uv_strerror(status));
```
dovholuknf, line 241: this could be pretty chatty. why not return an error status and propegate back through `on_ziti_data`? I imagine there's some reason, but it's not clear to me so i'm just gonna comment to bring your attention to the thought

library/conn_bridge.c
```diff
@@ -214,8 +307,36 @@ static void on_ziti_write(ziti_connection conn, ssize_t status, void *ctx) {
...
+        uv_timer_start(br->idle_timer, on_bridge_idle, br->idle_timeout, 0);
+    }
+
+    if (len > 0) {
+        ziti_write(br->conn, b->base, len, on_ziti_write, b->base);
+    } else {
+        pool_return_obj(b->base);
+        if (len == UV_ENOBUFS) {
+            if (!br->input_throttle) {
+                ZITI_LOG(TRACE, "stalled ziti_conn[%d.%d]", br->conn->ziti_ctx->id, br->conn->conn_id);
```
dovholuknf, line 323: something that "stalled" seems like it should be a warn? trace seems really low?

programs/ziti-prox-c/proxy.c
```diff
@@ -311,9 +312,18 @@ static void binding_listen_cb(ziti_connection srv, int status) {
 }
 
 static void on_ziti_accept(ziti_connection clt, int status) {
-    uv_stream_t *s = ziti_conn_data(clt);
+    uv_handle_t *s = ziti_conn_data(clt);
     if (status == ZITI_OK) {
-        ziti_conn_bridge(clt, s, on_bridge_close);
+        if (ziti_conn_bridge(clt, s, on_bridge_close) == 0) {
+            if (s->type == UV_UDP) {
+                ziti_conn_bridge_idle_timeout(clt, 10000);
```
dovholuknf, line 319: magic number `10000`? Should be a constant or configurable

dovholuknf (approved):
few questions/improvments but other than that lgtm

library/conn_bridge.c
ekoby: valid handles are specified in the doc

library/conn_bridge.c
ekoby: it won't be chatty because we close bridge on the first failure

library/conn_bridge.c
ekoby: it is not an error condition -- it just means that uplink(to ER) is slower than local connection -- which is true most of the time

programs/ziti-prox-c/proxy.c
ekoby: technically it is a literal constant

PR #514, opened 2023-05-02
properly deal with Windows paths

ekoby:


library/utils.c
```diff
@@ -509,7 +509,16 @@ static int parse_pkcs11_uri(const char *keyuri, tls_context *tls, void *ctx, par
     if (tlsuv_parse_url(&uri, keyuri) == 0) {
         if (uri.scheme_len == strlen("pkcs11") && strncmp(uri.scheme, "pkcs11", uri.scheme_len) == 0) {
             char lib[MAXPATHLEN] = "";
+#ifdef _WIN32
+            if (uri.path[0] == '/' && uri.path[2] == ':') {
```
dovholuknf, line 513: i feel like it would be better and more reliable to remove `pkcs11:///` and then take the rest of the string and verify the driver is found here instead of just an expected format.

dovholuknf: and a bonus is that code would also be cross-platform that way too. (determining if the driver is found)

ekoby: disagree, URL/URI parser already did its job. The `uri.path` looks like `/C:/...` on windows, and requires no manipulation on other platforms

ekoby: also removing three `/` is wrong for other platforms

dovholuknf: my point is that it sure would be friendly to tell the user at this time that the driver specified wasn't found so instead of just verifying the pattern, it would be better to verify the driver being specified actually exists.

as for `pkcs11://` vs `pkcs11:///`, if you're trying to emulate a file URI with this scheme, you need to treat the string like a file uri and look for the file at `file://` + `/c:/whatever` which is consistent with what "other OS's" and Windows expects

image

ekoby: that's how it is treated, is it not? the issue is that need to pass driver path to the code that does actual loading and checking that driver exists

dovholuknf (approved):

PR #515, opened 2023-05-05
allow logging into a file with default logger

ekoby:


library/utils.c
```diff
@@ -289,7 +289,16 @@ static void init_debug(uv_loop_t *loop) {
         tlsuv_set_debug(tlsuv_level, tlsuv_logger);
     }
 
-    ziti_debug_out = stderr;
+    char *logfile = getenv("ZITI_LOG_FILE");
+    if (logfile) {
```
dovholuknf, line 293: i don't love that this doesn't cover log rotation type policies. it also opens the file but never seems to close it? will that leak the descriptor when this process exits or will it get cleaned up?

library/utils.c
```diff
@@ -289,7 +289,16 @@ static void init_debug(uv_loop_t *loop) {
         tlsuv_set_debug(tlsuv_level, tlsuv_logger);
     }
 
-    ziti_debug_out = stderr;
+    char *logfile = getenv("ZITI_LOG_FILE");
```
scareything, line 292: is there a reason for doing this when the log level is being set? it feels like this belongs in ziti_log_init with explicit API around it.

library/utils.c
ekoby: logrotation belongs outside of SDK, this is just an option for default logger. An application using SDK is free to override it

ekoby: all process resources are cleanup up on exit

dovholuknf: why provide it at all then? the sdk already has callback based functionality. seems like it should either be here and complete, or not and we let the consumer decide if it logs to  a file

library/utils.c
ekoby: this is just an option for default logger, see comment below

library/utils.c
ekoby: it is probably the next step (this is how it is done in other libraries -- openssl, etc) 
I want to keep this for now until downstream projects can implement it -- probably needs zitilib layer API

PR #519, opened 2023-05-23
check if key supports certificate store

ekoby:


scareything (approved):
While looking into this I've seen stacks with `get_certificate` == NULL. Should we also check the `get_certificate` before calling it?

image

PR #526, opened 2023-06-14
Enhance parse_ziti_address_str CIDR prefix length parsing

tomc797:
Verify that `parse_ziti_address_str` prefix length parsing has made progress and that the prefix length is compatible with the address family. The behavior for `/ \0` (`slash` `nul`) has changed to assume maximum prefix length instead of zero. Moreover, we now ensure `hostname` is `nul` terminated. Silent `hostname` truncation behavior is maintained from the prior version. Lastly, the `addr` is zero initialized.

`parse_ziti_address_str` parses:
127.0.0.1/ -> 127.0.0.1/32
127.0.0.1/99 -> 127.0.0.1/32


ekoby:
we are moving towards supporting IDN hostname interpretations (see https://github.com/openziti/edge/pull/1518) which means that _invalid_ IP/CIDR-block addresses will be interpreted as host labels

tomc797:
Thanks for the information, @ekoby.

I've revised the PR to address your comment.

The fix does addresses two ```strncpy``` misuses and ensures that strings are ```nul``` terminated. It was noted that other parts of the code expected nul-terminated ```hostname```. See ```ziti_address_match``` for its use of ```strcasecmp```. This behavior is exhibited if ```addr_str``` is too long as demonstrated by:

```c
int
main(void)
{
    ziti_address *a1, *a2;
    void *m1, *m2;
    char *hostname;

    assert((m1 = malloc(4096)));
    assert((m2 = malloc(4096)));
    assert((hostname = malloc(258)));

    memset(m1, 'a', 4096);
    memset(m2, 'b', 4096);
    memset(hostname, 'c', 258); hostname[257] = '\0';

    a1 = m1;
    a2 = m2;
    assert(parse_ziti_address_str(a1, hostname) == 0);
    assert(parse_ziti_address_str(a2, hostname) == 0);

    int rc = ziti_address_match(a1, a2);
    printf("ziti_address_match(a1,a2) = %d\n", rc);
    printf("strncmp(a1->addr.hostname, a2->addr.hostname, %zu) = %d\n",
        sizeof a1->addr.hostname,
        strncmp(a1->addr.hostname, a2->addr.hostname, sizeof a1->addr.hostname));
    return 0;
}
```
producing
```
ziti_address_match(a1,a2) = -1
strncmp(a1->addr.hostname, a2->addr.hostname, 256) = 0
```
The PR gives:
```
ziti_address_match(a1,a2) = 0
strncmp(a1->addr.hostname, a2->addr.hostname, 256) = 0
```
As mentioned above, invalid CIDRs are now mapped to hostnames. For example,
```c
void
print(const ziti_address *a)
{
  if (a->type == ziti_address_hostname) {
    printf("hostname: %s\n", a->addr.hostname);
  } else if (a->type == ziti_address_cidr) {
    printf("cidr: %s/%u\n",
        inet_ntoa((struct in_addr){ .s_addr = a->addr.cidr.ip.s6_addr32[0] }),
        a->addr.cidr.bits);
  } else {
    printf("invalid\n");
  }
}

int
main(void)
{
    ziti_address a1, a2, a3, a4;

    assert(parse_ziti_address_str(&a1, "0.0.0.0/0") == 0);
    assert(parse_ziti_address_str(&a2, "100.64.0.1/10") == 0);
    assert(parse_ziti_address_str(&a3, "100.64.0.1/32") == 0);
    print(&a1);
    print(&a2);
    print(&a3);

    assert(parse_ziti_address_str(&a1, "100.64.0.1/") == 0);
    assert(parse_ziti_address_str(&a2, "100.64.0.1/       47") == 0);
    assert(parse_ziti_address_str(&a3, "100.64.0.1/-2") == 0);
    assert(parse_ziti_address_str(&a4, "100.64.0.1/18446744073709551615") == 0);
    print(&a1);
    print(&a2);
    print(&a3);
    print(&a4);
    return 0;
}
```
Prior to the PR, the code produces:
```
cidr: 0.0.0.0/0
cidr: 100.64.0.1/10
cidr: 100.64.0.1/32
cidr: 100.64.0.1/0
cidr: 100.64.0.1/47
cidr: 100.64.0.1/4294967294
cidr: 100.64.0.1/4294967295
```
Note the large prefix length. Prefix length is often used in computing the mask; large numbers could trigger undefined behavior due to negative shift value. After the PR:
```
cidr: 0.0.0.0/0
cidr: 100.64.0.1/10
cidr: 100.64.0.1/32
hostname: 100.64.0.1/
hostname: 100.64.0.1/       47
hostname: 100.64.0.1/-2
hostname: 100.64.0.1/18446744073709551615
```

ekoby:
@tomc797 Tom, I am ready to merge it but I would like to add you to CONTRIBUTORS file. Let me know if you prefer not to.

ekoby (approved):

tomc797:
Thanks, @ekoby . I've added my name to CONTRIBUTORS.

PR #534, opened 2023-06-26
add CONTRIBUTORS file

ekoby:


dovholuknf (approved):

CONTRIBUTORS
```diff
@@ -0,0 +1,2 @@
+# Authors ordered by first contribution
```
smilindave26, line 1: How does this compare/keep-up-to-date with https://github.com/openziti/ziti-sdk-c/graphs/contributors?

ekoby: I don't think it is not related to that at all. 
We would just ask them to add themselves to the file on their first PR, they can opt out if they don't feel like it

PR #540, opened 2023-07-21
put outbound close messages on the write queue

### LINKED ISSUE #539: ziti_close causes queued write requests to be dropped
`ziti_close` immediately sends a `ContentTypeStateClosed` message, bypassing the connection's write queue. Any messages that are on the queue when `ziti_close` is called will not be delivered.

scareything:
fixes #539

### DIFF
--- a/library/connect.c
+++ b/library/connect.c
@@ -573,6 +573,12 @@ static void ziti_write_req(struct ziti_write_req_s *req) {
         conn->write_reqs--;
         free(req);
         return;
+    } else if (req->close) {
+        // conn->state will be set on_disconnect callback
+        message *m = create_message(conn, ContentTypeStateClosed, 0);
+        send_message(conn, m, req);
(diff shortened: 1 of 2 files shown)


library/connect.c
```diff
@@ -615,12 +621,11 @@ static void ziti_disconnect_async(struct ziti_conn *conn) {
         case Connected:
         case CloseWrite:
         case Timedout: {
-            message *m = create_message(conn, ContentTypeStateClosed, 0);
             NEWP(wr, struct ziti_write_req_s);
             wr->conn = conn;
             wr->cb = on_disconnect;
```
ekoby, line 626: shouldn't `wr->close` be set here?

scareything: thanks. I swear it was being set at one point.

ekoby (approved):

PR #542, opened 2023-08-07
used versioned builder image

qrkourier:


ekoby (approved):
any way to specify version once?

qrkourier:
> any way to specify version once?

Shouldn't be a problem to do that.

PR #557, opened 2023-09-26
Avoid crash when app omits config

scareything:


ekoby (approved):

scareything:
gave up due to inexplicable unsigned commits appearing in this branch after rebasing

PR #559, opened 2023-09-29
Implement saner init

ekoby:


programs/ziti-prox-c/proxy.c
```diff
@@ -590,6 +591,12 @@ void run(int argc, char **argv) {
...
     }
 
+    ziti_config cfg;
+    ziti_context ztx;
+
+    ziti_load_config(&cfg, config);
+    ziti_context_init(&ztx, &cfg);
+
     ziti_options opts = {
             .config = config,
```
scareything, line 601: should the config string be present in the opts if the config is being passed as an object through the context?

library/ziti.c
```diff
@@ -1783,3 +1796,93 @@ void ziti_queue_work(ziti_context ztx, ztx_work_f w, void *data) {
...
+    *ztx = ctx;
+    return ZITI_OK;
+}
+
+
+int ziti_context_set_options(ziti_context ztx, const ziti_options *options) {
+    if (options == NULL) {
+        ztx->opts = default_options;
+    } else {
+#define copy_opt(f) if (options->f != 0) ztx->opts.f = options->f
```
scareything, line 1841: what's the point of this macro? I'd think it would be OK for the 0 values to be assigned as well?

library/ziti.c
```diff
@@ -1783,3 +1796,93 @@ void ziti_queue_work(ziti_context ztx, ztx_work_f w, void *data) {
...
+
+int ziti_context_init(ziti_context *ztx, const ziti_config *config) {
+    if (config == NULL || config->controller_url == NULL) {
+        ZITI_LOG(ERROR, "config or controller/tls has to be set");
+        return ZITI_INVALID_CONFIG;
+    }
+
+    ziti_context ctx = calloc(1, sizeof(*ctx));
+
+    if (strncmp(config->id.ca, "file://", strlen("file://")) == 0) {
```
scareything, line 1808: I think we need to test `config->id.ca` for null before passing it to `strncmp`.

library/ziti.c
```diff
@@ -1783,3 +1796,93 @@ void ziti_queue_work(ziti_context ztx, ztx_work_f w, void *data) {
...
+        struct tlsuv_url_s url;
+        if (tlsuv_parse_url(&url, config->id.ca) != 0) {
```
scareything, line 1810: null check also needed here

library/ziti.c
```diff
@@ -1783,3 +1796,93 @@ void ziti_queue_work(ziti_context ztx, ztx_work_f w, void *data) {
...
+            ZITI_LOG(ERROR, "invalid CA bundle reference[]");
+            return ZITI_INVALID_CONFIG;
+        }
+
+        char *ca = NULL;
+        size_t ca_len;
+        int rc = load_file(url.path, url.path_len, &ca, &ca_len);
+        if (rc == 0) {
+            FREE(ctx->config.id.ca);
```
scareything, line 1819: this seems suspect. the caller may be responsible for this pointer?

programs/ziti-prox-c/proxy.c
ekoby: it will be ignored anyway, and field will be removed in the future

library/ziti.c
ekoby: the goal is to allow partial updates, without accidentally whacking values set from `default_options`
for example, the app may not care about the `api_page_size` so it won't set it in the struct.

library/ziti.c
ekoby: all passed structs are `const` and their content is copied into `ziti_context`. this avoids ownership ambiguity

scareything:
Since this is an example, would it make sense to remove it here?On Sep 29, 2023, at 5:39 PM, ekoby ***@***.***> wrote:﻿
@ekoby commented on this pull request.



In programs/ziti-prox-c/proxy.c:
> @@ -590,6 +591,12 @@ void run(int argc, char **argv) {
         model_map_set(&app_ctx.listeners, service_name, l);
     }
 
+    ziti_config cfg;
+    ziti_context ztx;
+
+    ziti_load_config(&cfg, config);
+    ziti_context_init(&ztx, &cfg);
+
     ziti_options opts = {
             .config = config,

it will be ignored anyway, and field will be removed in the future

programs/ziti-prox-c/proxy.c
scareything: Since this app serves as a clear example, would it make sense to remove it here?

scareything (approved):

PR #563, opened 2023-10-12
fix: remove_binding() was called twice

ekoby:
fix: remove_binding() was called twice if channel close event happened in the middle of bind operation

### DIFF
--- a/library/bind.c
+++ b/library/bind.c
@@ -33,6 +33,7 @@ struct binding_s {
     ziti_channel_t *ch;
     struct key_pair key_pair;
     bool bound;
+    struct waiter_s *waiter;
 };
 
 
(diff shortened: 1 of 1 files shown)


library/bind.c
```diff
@@ -453,6 +465,7 @@ static void remove_binding(struct binding_s *b) {
 void on_unbind(void *ctx, message *m, int code) {
     struct binding_s *b = ctx;
     struct ziti_conn *conn = b->conn;
+    b->waiter = NULL;
```
scareything, line 468: did this waiter get removed from `b->ch->waiters`?

scareything (approved):

library/bind.c
ekoby: yes, once you get a reply callback the waiter is no longer valid

PR #574, opened 2023-11-15
fix crash on rebind

### LINKED ISSUE #570: services are not bound, and assertion fails when edge router flaps
1. create two identities and a service that each identity binds
2. start clients, wait for both to successfully bind:

       ziti-prox-c run -c id1.json -b testservice:127.0.0.1:1234 &
       ziti-prox-c run -c id2.json -b testservice:127.0.0.1:1234 &

3. stop the edge router. note that terminators remain even though router is down and services are not bound.
4. restart router. note that terminators are removed and not recreated.

The clients do not re-bind even after the next re-bind attempt (300s) after the router was stopped. Additionally, the clients hit this assert on the next re-bind attempt:

```
(23302)[      315.855]   DEBUG ziti-sdk:bind.c:96 rebind_delay_cb() server[0.0] staring re-bind
(23302)[      315.907]   DEBUG ziti-sdk:ziti_ctrl.c:329 ctrl_body_cb() ctrl[shawns-m1-mbp.localdomain] completed GET[/sessions/clot4lqgh2j4c63ff5vlb53rb] in 0.051 s
(23302)[      315.907]   DEBUG ziti-sdk:bind.c:124 process_bindings() server[0.0] checking er.shawns-m1-mbp[tls://shawns-m1-mbp.localdomain:3022]
Assertion failed: (ch == b->ch), function process_bindings, file bind.c, line 136.
```

edit: I'm not sure if it's relevant but the network that I recreated this on had two edge routers, only one of which was online (and restarted) for the duration of the test.

edit2: a single enrolled router is sufficient to reproduce the issue. controller and router 0.31.0.

ekoby:
- keep binding objects around until server close
[fixes #570]

### DIFF
--- a/library/bind.c
+++ b/library/bind.c
@@ -45,11 +45,9 @@ static void get_service_cb(ziti_context, ziti_service *service, int status, void
 
 static int dispose(ziti_connection server);
 
-static void start_binding(struct binding_s *b);
+static void start_binding(struct binding_s *b, ziti_channel_t *ch);
 
 static void stop_binding(struct binding_s *b);
-
-static void remove_binding(struct binding_s *b);
 
 static void schedule_rebind(struct ziti_conn *conn, bool now);
 
@@ -102,10 +100,9 @@ static void rebind_delay_cb(uv_timer_t *t) {
     }
 }
 
-static struct binding_s* new_binding(struct ziti_conn *conn, ziti_channel_t *ch) {
+static struct binding_s* new_binding(struct ziti_conn *conn) {
     NEWP(b, struct binding_s);
     b->conn = conn;
-    b->ch = ch;
     init_key_pair(&b->key_pair);
     return b;
 }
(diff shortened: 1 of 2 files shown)


library/bind.c
```diff
@@ -265,6 +261,17 @@ static uint8_t get_terminator_precedence(const ziti_listen_opts *opts, const cha
 static int dispose(ziti_connection server) {
     assert(server->type == Server);
 
+    model_map_iter it = model_map_iterator(&server->server.bindings);
+    while(it) {
+        struct binding_s *b = model_map_it_value(it);
+        if (!b->bound && b->waiter == NULL) {
+            it = model_map_it_remove(it);
+            free(b);
```
dovholuknf, line 269: don't need to free b->bound or b->waiter, right? i assume not but always better to ask i suppose :)

dovholuknf (approved):
other than the small question, which i'm sure is fine.. lgtm

library/bind.c
ekoby: waiters are owned by the channel

PR #575, opened 2023-11-17
add ListenerId header in bind/unbind requests

ekoby:


library/bind.c
```diff
@@ -66,6 +66,7 @@ int ziti_bind(ziti_connection conn, const char *service, const ziti_listen_opts
     conn->type = Server;
     conn->disposer = dispose;
     conn->service = strdup(service);
+    uv_random(NULL, NULL, conn->server.listener_id, sizeof(conn->server.listener_id), 0 , NULL);
```
scareything, line 69: should this be an async call with a timeout, and also some error checking/handling here? [uv_random(http://docs.libuv.org/en/v1.x/misc.html#c.uv_random):
- fails if it can't produce the requested number of random bytes
- may "block indefinitely" (synchronous call), or "never finish" (async call) when there is insufficient entropy.

library/bind.c
```diff
@@ -518,9 +524,15 @@ static void stop_binding(struct binding_s *b) {
                     .length = sizeof(conn_id),
                     .value = (uint8_t *) &conn_id
             },
+            {
+                    .header_id = ListenerId,
+                    .length = sizeof(b->conn->server.listener_id),
+                    .value = (uint8_t*)b->conn->server.listener_id,
```
scareything, line 530: we're sending raw bytes over the wire. should the value be normalized for different endian in some way?

library/bind.c
ekoby: we're only asking for 32 bytes here, it should not be an issue

library/bind.c
ekoby: the bites are not interpreted is any way and are used for comparisons only

scareything (approved):

PR #584, opened 2023-12-09
more changes for tlsuv 0.26

scareything:


scareything:
rebasing from 0.35.5 instead

PR #601, opened 2023-12-20
Ziti dump connection stats

ekoby:
add connection marker in ziti log and dump
new data in dump:
- marker
- time to connect
- idle time
- bytes sent/received
- underlay info if connection is bridged

sample dump output:
```
conn[1/hTz4t7n2]: state[Connected] service[test1] using ch[0/quickstart-router]
	connect_time[64] idle_time[8671ms] sent[21] recv[0] recv_buff[0]
	bridge: tcp: 127.0.0.1:18000 -> 127.0.0.1:52180
conn[0]: server service[test1] terminators[1]
	child[2/hTz4t7n2]: state[Connected] caller_id[tester] ch[0/quickstart-router]
		accept_time[1] idle_time[8670ms] sent[0] recv[21] recv_buff[0]
	        bridge: tcp: 127.0.0.1:52182 -> 127.0.0.1:19000
```

### DIFF
--- a/library/conn_bridge.c
+++ b/library/conn_bridge.c
@@ -53,6 +53,8 @@ static void close_bridge(struct ziti_bridge_s *br);
 
 static void on_input(uv_stream_t *s, ssize_t len, const uv_buf_t *b);
 static void on_udp_input(uv_udp_t *udp, ssize_t len, const uv_buf_t *b, const struct sockaddr *addr, unsigned int flags);
+static int fmt_addr(struct sockaddr_storage *ss, char *host, size_t host_len, int *port);
+
 
 extern int ziti_conn_bridge(ziti_connection conn, uv_handle_t *handle, uv_close_cb on_close) {
     if (handle == NULL) return UV_EINVAL;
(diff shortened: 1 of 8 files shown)


library/ziti.c
```diff
@@ -622,6 +622,7 @@ int ziti_get_appdata(ziti_context ztx, const char *key, void *data,
 
 
 void ziti_dump(ziti_context ztx, int (*printer)(void *arg, const char *fmt, ...), void *ctx) {
+    uint64_t now = uv_now(ztx->loop);
```
dovholuknf, line 625: why so far from where it's actually used?

library/connect.c
```diff
@@ -515,6 +519,11 @@ static int do_ziti_dial(ziti_connection conn, const char *service, ziti_dial_opt
         return ZITI_INVALID_STATE;
     }
 
+    char marker[6];
```
dovholuknf, line 522: 6 really should be a defined constant somewhere imo

library/connect.c
```diff
@@ -848,6 +861,7 @@ void conn_inbound_data_msg(ziti_connection conn, message *msg) {
                 CONN_LOG(VERBOSE, "decrypted %lld bytes", plain_len);
                 buffer_append(conn->inbound, plain_text, plain_len);
                 metrics_rate_update(&conn->ziti_ctx->down_rate, (int64_t) plain_len);
+                conn->received += plain_len;
```
dovholuknf, line 864: shouldn't this be in the "non-encrypted" section too (like 876 ish)?

library/conn_bridge.c
```diff
@@ -412,3 +416,71 @@ void on_input(uv_stream_t *s, ssize_t len, const uv_buf_t *b) {
         }
     }
 }
+
+int conn_bridge_info(ziti_connection conn, char *buf, size_t buflen) {
+    if (conn == NULL || !conn->bridged) {
+        return ZITI_INVALID_STATE;
```
dovholuknf, line 422: does this deserve a log message?

ekoby: no, it is an internal information function -- this return code just means the connection is not bridged

dovholuknf (approved):

PR #603, opened 2023-12-21
fix loading ziti_config from in-memory JSON (fixes #602)

### LINKED ISSUE #602: Ziti_load_context does not process the pointer as json
@ekoby informs me that `Ziti_load_context` is intended to process the pointer as either a file on disk or json. Using it for reading an identity file works, however using the pointer as json does not.

Please update the zitilib.h file and clarify that this pointer is either json or a file on disk. Also indicate the precedence (is it looking for a file first, then json, or json first then a file?). Also indicate if the path is absolute only or if it can be relative and finally, fix the function to take either a pointer to json or a pointer that represents a path on the file system

ekoby:


includes/ziti/ziti.h
```diff
@@ -420,8 +420,16 @@ extern int ziti_enroll(ziti_enroll_opts *opts, uv_loop_t *loop, ziti_enroll_cb e
 ZITI_FUNC
 extern void ziti_set_app_info(const char *app_id, const char *app_version);
 
+/**
+ * Load ziti identity config from memory or file.
+ * First it tries to parse [conf_str] as identity Json.
+ * if that fails it tries to load it from file using [conf_str] as the path.
```
dovholuknf, line 426: does it allow for relative path from cwd too?

library/config.c
```diff
@@ -60,8 +60,12 @@ int ziti_load_config(ziti_config *cfg, const char* cfgstr) {
     }
 
     if (rc < 0) {
-        free_ziti_config(cfg);
+        return rc;
```
dovholuknf, line 63: should free_ziti_config still be called?

includes/ziti/ziti.h
ekoby: yes

library/config.c
dovholuknf: and actually, shouldn't this whole block just move up to 57-60? it's the same if statement

dovholuknf: oooh i see. it's testing it twice and setting it ... nevermind ☝️

ekoby: no, `rc` changes between 57 and here

dovholuknf (approved):

PR #604, opened 2023-12-29
Cleanup connection timers

ekoby:


library/connect.c
```diff
@@ -1063,19 +1016,19 @@ static int ziti_channel_start_connection(struct ziti_conn *conn, ziti_channel_t
...
-            nheaders++;
-        }
-        if (req->dial_opts->app_data != NULL) {
-            headers[nheaders].header_id = AppDataHeader;
-            headers[nheaders].value = req->dial_opts->app_data;
-            headers[nheaders].length = req->dial_opts->app_data_sz;
-            nheaders++;
-        }
+
+    if (req->dial_opts.identity != NULL) {
```
scareything, line 1020: Might be worth checking for empty string here too. I see a lot of configs with dial / listen options set to `""`.

scareything (approved):

library/connect.c
ekoby: I added the check in clone_dial_opts

PR #608, opened 2024-01-04
do not log message if channel closed

ekoby:


library/bind.c
```diff
@@ -405,8 +405,8 @@ static void bind_reply_cb(void *ctx, message *msg, int code) {
     struct ziti_conn *conn = b->conn;
 
     b->waiter = NULL;
-    CONN_LOG(TRACE, "received msg ct[%X] code[%d]", msg->header.content, code);
     if (code == ZITI_OK && msg->header.content == ContentTypeStateConnected) {
```
scareything, line 408: can `msg` be null when code == `ZITI_OK`?

scareything (approved):

library/bind.c
ekoby: no

PR #615, opened 2024-01-30
Channel complete close before reconnect

ekoby:


library/channel.c
```diff
@@ -120,13 +120,19 @@ struct msg_receiver {
...
 
 static void ch_init_stream(ziti_channel_t *ch) {
-    if (ch->connection == NULL) {
-        ch->connection = calloc(1, sizeof(*ch->connection));
-        tlsuv_stream_init(ch->loop, ch->connection, ch->ctx->tlsCtx);
-        tlsuv_stream_keepalive(ch->connection, true, 30);
-        tlsuv_stream_nodelay(ch->connection, true);
-        ch->connection->data = ch;
+
+    if (ch->connection != NULL) {
```
scareything, line 124: ch->connection will be leaked at line 130

ekoby: this is a leftover from debugging. this function should not be called when `ch->connection != NULL`

library/channel.c
```diff
@@ -210,8 +216,15 @@ int ziti_close_channels(struct ziti_ctx *ztx, int err) {
 
 static void on_tls_close(uv_handle_t *s) {
     tlsuv_stream_t *tls = (tlsuv_stream_t *) s;
+    ziti_channel_t *ch = tls->data;
+    ch->connection = NULL;
```
scareything, line 220: is there another reference to the connection? this looks like another leak.

ekoby: handle(`tls`) is the same as `ch->connection`

scareything (approved):

PR #624, opened 2024-02-20
add a freebsd preset

qrkourier:


CMakePresets.json
```diff
@@ -167,6 +167,22 @@
...
+        "ci-unix",
+        "dev-mode",
+        "vcpkg"
+      ],
+      "cacheVariables": {
+        "CMAKE_MAKE_PROGRAM": "ninja",
+        "CMAKE_C_COMPILER": "clang",
+        "CMAKE_CXX_COMPILER": "clang++",
+        "VCPKG_TARGET_TRIPLET": "x64-freebsd"
+        // "VCPKG_CHAINLOAD_TOOLCHAIN_FILE": "${sourceDir}/toolchains/FreeBSD-x64.cmake"
```
ekoby, line 183: can't have comments in JSON

qrkourier: That's true, and it looks like it's only working this way because of [a known issue](https://gitlab.kitware.com/cmake/cmake/-/issues/22602). Perhaps CMake will begin to support YAML so we can have comments.

That may never happen because [other things need to read the presets JSON file](https://discourse.cmake.org/t/future-yaml-support-for-cmake-presets/7188/5), not just CMake.

PR #631, opened 2024-03-05
Misc refactorings

ekoby:


scareything (approved):
will this come with a new minor number?

ekoby:
> will this come with a new minor number?

no, there is no external API changes

PR #655, opened 2024-05-31
More flexible versions

ekoby:


cmake/version.cmake
```diff
@@ -1,48 +1,49 @@
-
-function(get_version version_file version_var branch_var commit_var)
+macro(do_git out_var err_var)
     execute_process(
-            COMMAND git describe --tags --match=[0-9]* HEAD
+            COMMAND "git" ${ARGN}
```
dovholuknf, line 3: this work on windows? before it used git commands but this looks like it uses a bash variable and that might change things?

dovholuknf (approved):

PR #672, opened 2024-06-13
add mfa status cb to auth_method

ekoby:


qrkourier (approved):
works on my machine (fix for unresponsive IPC mfacode)

PR #675, opened 2024-06-27
avoid passing incompatible pointer types to message_get_bytes_header

scareything:
fixes builds on gcc 14.

### DIFF
--- a/inc_internal/message.h
+++ b/inc_internal/message.h
@@ -90,7 +90,7 @@ bool message_get_int32_header(message *m, int header_id, int32_t *v);
 
 bool message_get_uint64_header(message *m, int header_id, uint64_t *v);
 
-bool message_get_bytes_header(message *m, int header_id, uint8_t **ptr, size_t *len);
+bool message_get_bytes_header(message *m, int header_id, const uint8_t **ptr, size_t *len);
 
 uint8_t *write_hdr(const hdr_t *h, uint8_t *buf);
 
--- a/inc_internal/message.h
+++ b/inc_internal/message.h
@@ -90,7 +90,7 @@ bool message_get_int32_header(message *m, int header_id, int32_t *v);
 
 bool message_get_uint64_header(message *m, int header_id, uint64_t *v);
 
-bool message_get_bytes_header(message *m, int header_id, const char **ptr, size_t *len);
+bool message_get_bytes_header(message *m, int header_id, uint8_t **ptr, size_t *len);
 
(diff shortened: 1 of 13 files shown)


dovholuknf (approved):

inc_internal/message.h
```diff
@@ -90,7 +90,7 @@ bool message_get_int32_header(message *m, int header_id, int32_t *v);
 
 bool message_get_uint64_header(message *m, int header_id, uint64_t *v);
 
-bool message_get_bytes_header(message *m, int header_id, const char **ptr, size_t *len);
+bool message_get_bytes_header(message *m, int header_id, uint8_t **ptr, size_t *len);
```
ekoby, line 93: that is a bad idea -- this returns the pointer to data owned by the message -- therefore marked const to the caller.
The caller cannot assume that the data is modifiable.

library/bind.c
```diff
@@ -381,7 +381,7 @@ static void process_dial(struct binding_s *b, message *msg) {
     bool caller_id_sent = message_get_bytes_header(msg, CallerIdHeader, &source_identity, &source_identity_sz);
 
     ziti_client_ctx clt_ctx = {0};
-    message_get_bytes_header(msg, AppDataHeader, (uint8_t **) &clt_ctx.app_data, &clt_ctx.app_data_sz);
+    message_get_bytes_header(msg, AppDataHeader, &clt_ctx.app_data, &clt_ctx.app_data_sz);
```
ekoby, line 384: it would be safer to make a copy here -- otherwise app_data points to garbage once msg is discarded

ekoby (changes requested):
current `app_data` header usage seems unsafe

library/bind.c
scareything: Just to be clear, the variable scoping has not been changed in this PR. 

I believe `cit_ctx->app_data` will be assigned the pointer from the message header. `clt_ctx` is a stack variable, which is only passed to the client callback. We do [document that the client context is only valid while the callback is on the stack](https://github.com/openziti/ziti-sdk-c/blob/main/includes/ziti/ziti.h#L337).

inc_internal/message.h
scareything: changed signature back to const.

library/bind.c
ekoby: I think to make it more clear both `clt_ctx` and `ctl_ctx.app_data` should be `const` in the callback

ekoby (approved):

PR #677, opened 2024-06-27
fix parse_ref handling of file:/, file://, file:///

dovholuknf:


library/ziti.c
```diff
@@ -96,7 +96,10 @@ static size_t parse_ref(const char *val, const char **res) {
     if (val != NULL) {
         if (strncmp("file:", val, 5) == 0) {
             // load file
-            *res = val + strlen("file://");
+            struct tlsuv_url_s url;
+            tlsuv_parse_url(&url, val);
+            size_t start = strlen(val) - strlen(url.path);
+            *res = val + start;
```
ekoby, line 102: isn't this just `*res = url.path`?

dovholuknf: wouldn't that pointer (url.path) go out of scope as soon as the function returns? it would need to use val, no?

library/ziti.c
```diff
@@ -96,7 +96,10 @@ static size_t parse_ref(const char *val, const char **res) {
...
             len = strlen(*res) + 1;
```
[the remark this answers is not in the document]
dovholuknf: that seems reasonable, sure.

library/ziti.c
dovholuknf: actually yeah i see what you mean now i thought about it a bit more.

scareything (approved):
I don’t think there’s a downside to using macOS-latest here.

dovholuknf:
another PR that i thinjk should go to releaes 1.x... i'm closing this and will make these same changes over on https://github.com/openziti/ziti-sdk-c/pull/680

PR #685, opened 2024-07-09
Bump integ test

ekoby:
disable integ-test failure for now


qrkourier (approved):
Do we want dependabot to auto-update the CLI version? I'm unsure if there's a clearly-defined way to do that, but will explore if that'd be desirable.

scareything (approved):

ekoby:
> Do we want dependabot to auto-update the CLI version? I'm unsure if there's a clearly-defined way to do that, but will explore if that'd be desirable.

we can use major version for the pin

PR #710, opened 2024-08-16
Model improvement

ekoby:
- rename string => model_string -- avoid collision in C++
- rename int => number and extend to 64 bit


dovholuknf:
why not do them all and prefix them all with `model_`? like `model_int` instead of `number`? `model_bool` etc?

dovholuknf (approved):

PR #713, opened 2024-08-19
better ctrl response parsing

ekoby:
- parse JSON as it comes it
- avoid double-parsing of response data

### DIFF
--- a/includes/ziti/ziti_model.h
+++ b/includes/ziti/ziti_model.h
@@ -214,7 +214,7 @@ ZITI_FUNC int ziti_address_print(char *buf, size_t max, const ziti_address *addr
 
 ZITI_FUNC int ziti_address_match(const ziti_address *addr, const ziti_address *range);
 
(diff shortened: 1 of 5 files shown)


includes/ziti/ziti_model.h
```diff
@@ -214,7 +214,7 @@ ZITI_FUNC int ziti_address_print(char *buf, size_t max, const ziti_address *addr
 
 ZITI_FUNC int ziti_address_match(const ziti_address *addr, const ziti_address *range);
 
-ZITI_FUNC int ziti_address_match_s(const char *addr, ziti_address *range);
+ZITI_FUNC int ziti_address_match_s(const char *addr, const ziti_address *range);
```
dovholuknf, line 217: minor version bump?

ekoby: it will go into 2.0-alpha so minor bumps are not needed yet

PR #714, opened 2024-08-19
allow application to force service refresh

ekoby:
- appropriate events will be triggered
- allows mobile clients to reduce polling and still present up-to-date info to the user

### DIFF
--- a/includes/ziti/ziti.h
+++ b/includes/ziti/ziti.h
@@ -444,11 +444,11 @@ extern int ziti_context_run(ziti_context ztx, uv_loop_t *loop);
  * \brief Trigger refresh ahead of normal refresh cycle.
  *
  * This method will force ziti context to update its internal model from ziti controller.
- * Appropriate events will be triggers as needed
- * (just like during normal refresh cycle [ziti_options.refresh_interval]):
+ * Appropriate events will be triggered as needed
+ * (just like during normal refresh cycle [#ziti_options.refresh_interval]):
  * - Service Added/Removed, etc.
  * @param ztx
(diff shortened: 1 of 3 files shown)


includes/ziti/ziti.h
```diff
@@ -440,6 +440,19 @@ extern int ziti_context_set_options(ziti_context ztx, const ziti_options *option
...
 
+/**
+ * \brief Trigger refresh ahead of normal refresh cycle.
+ *
+ * This method will force ziti context to update its internal model from ziti controller.
+ * Appropriate events will be triggers as needed
+ * (just like during normal refresh cycle [ziti_options.refresh_interval]):
+ * - Service Added/Removed, etc.
+ * @param ztx
+ * @return 0 on success, or error code
```
dovholuknf, line 451: seems like `0` should be `ZITI_OK`.

dovholuknf: typo -> "will be triggered as needed"

library/ziti.c
```diff
@@ -1613,6 +1613,15 @@ int ziti_context_run(ziti_context ztx, uv_loop_t *loop) {
     return ZITI_OK;
 }
 
+int ziti_refresh(ziti_context ztx) {
+    if (!ztx->enabled) return ZITI_DISABLED;
+
+    ZTX_LOG(INFO, "application requested service refresh");
```
dovholuknf, line 1619: feels like a DEBUG to me or even maybe removing it. i called it -- i could log it?

dovholuknf (approved):

PR #718, opened 2024-08-29
Implement ext JWT authentication for HA

ekoby:
add access to available external authentication providers


includes/ziti/ziti.h
```diff
@@ -509,6 +509,15 @@ extern const char *ziti_get_controller(ziti_context ztx);
 ZITI_FUNC
 extern const ziti_identity *ziti_get_identity(ziti_context ztx);
 
+
+typedef void (*ziti_ext_signers_cb)(ziti_context, int err, ziti_jwt_signer_array, void*);
+
+ZITI_FUNC
+extern int ziti_get_ext_jwt_signers(ziti_context ztx, ziti_ext_signers_cb, void *ctx);
```
dovholuknf, line 516: `ziti.h` should have some sort of doc with it

library/oidc.c
```diff
@@ -417,6 +427,30 @@ static void ext_done(uv_work_t *wr, int status) {
 }
 
 static void start_ext_auth(auth_req *req, const char *ep, int qc, tlsuv_http_pair q[]) {
+    uv_loop_t *loop = req->clt->timer->loop;
+    struct sockaddr_in6 addr = {
```
dovholuknf, line 431: what happens when the OS doesn't support ipv6 or has it disabled? lots of the internet was recommending disabling ipv6 for a long time, i'm just wondering if this will be hard to understand/debug and if we can give the user a decent error

dovholuknf (approved):
add doc for ziti.h, pls

PR #736, opened 2024-10-01
implement more robust ziti_channel re-connect

ekoby:
fix: failed write put ziti_channel into partially disconnected state

### DIFF
--- a/library/channel.c
+++ b/library/channel.c
@@ -226,8 +226,6 @@ int ziti_channel_close(ziti_channel_t *ch, int err) {
         uv_close((uv_handle_t *) ch->timer, (uv_close_cb) free);
         ch->timer = NULL;
 
-        close_connection(ch);
-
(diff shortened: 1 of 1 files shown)


library/channel.c
```diff
@@ -918,8 +914,23 @@ static void on_channel_data(uv_stream_t *s, ssize_t len, const uv_buf_t *buf) {
...
+
+    // connect request was cancelled via tlsuv_stream_close
+    // cleanup in close callback
+    if (status == UV_ECANCELED || tls->data == NULL) {
+        goto done;
+    }
+
     ziti_channel_t *ch = tls->data;
 
+    if (tls != ch->connection) {
```
scareything, line 926: should `ch` still be checked for null?

ekoby: not really necessary, we always set it to `ch`

scareything (approved):

PR #746, opened 2024-10-09
add another call to try to closeWindow

dovholuknf:
seems to work for me on windows+brave. maybe i'll try edge too...

seem-to-work


dovholuknf:
worked with edge and chrome too but only if there was already an open window and this was a new tab. when it starts the browser up on its own, it doesn't seem to close the "final" window

ekoby (approved):

PR #747, opened 2024-10-09
main --> release-1.x

scareything:


scareything:
we will make a new branch for stable releases instead of continuing release-1.x.

PR #756, opened 2024-10-17
attempt to suppress the WARN thatpops when loading an identity as a file

dovholuknf:


library/config.c
```diff
@@ -50,12 +50,25 @@ int ziti_load_config(ziti_config *cfg, const char* cfgstr) {
     if (!cfgstr) {
         return ZITI_INVALID_CONFIG;
     }
+    bool seems_like_json = false;
+    const char* c = cfgstr;
+    while (*cfgstr && isspace((unsigned char)*cfgstr)) {
+        c++;
+    }
+    if (strncmp(c,"{",1) == 0) {
```
ekoby, line 58: `if (*c  == '{')` ??

dovholuknf: yep. done

library/config.c
```diff
@@ -50,12 +50,25 @@ int ziti_load_config(ziti_config *cfg, const char* cfgstr) {
     if (!cfgstr) {
         return ZITI_INVALID_CONFIG;
     }
+    bool seems_like_json = false;
+    const char* c = cfgstr;
+    while (*cfgstr && isspace((unsigned char)*cfgstr)) {
```
ekoby, line 55: this should check `*c` not `cfgstr` otherwise it is infinite loop

dovholuknf: yeah, just so happened that the first char wasn't whitespace when i tested it. nice catch

library/config.c
```diff
@@ -50,12 +51,25 @@ int ziti_load_config(ziti_config *cfg, const char* cfgstr) {
...
+        c++;
+    }
+    if (*c == '{') {
+        seems_like_json = true;
+    }
 
     memset(cfg, 0, sizeof(*cfg));
-    int rc = parse_ziti_config(cfg, cfgstr, strlen(cfgstr));
+    int rc;
+    if(seems_like_json) {
```
scareything, line 65: whitespace

ekoby (approved):

PR #759, opened 2024-10-22
don't list routers until session is available

scareything:
list_routers_cb calls process_bindings, which needs the session


scareything:
superseded by #760

PR #761, opened 2024-10-23
check for disabled ztx before using ztx_get_controller

scareything:


smilindave26 (approved):
would a better message be "ignored - posture checked are disabled"?

PR #772, opened 2024-11-20
Optimize bind refresh

### LINKED ISSUE #769: avoid superfluous service binding refreshes
unnecessary refreshed cause a lot traffic which increase costs in certain environments (cellular/satellite networks)

Notes: make refresh happen on certain events instead of timer

ekoby:
[fixes #769]

### DIFF
--- a/inc_internal/message.h
+++ b/inc_internal/message.h
@@ -55,6 +55,9 @@ typedef struct {
     uint8_t *value;
 } hdr_t;
 
+#define var_header(id, var) header(id, sizeof(var), &(var))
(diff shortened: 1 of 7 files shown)


inc_internal/message.h
```diff
@@ -55,6 +55,9 @@ typedef struct {
     uint8_t *value;
 } hdr_t;
 
+#define var_header(id, var) header(id, sizeof(var), &(var))
```
scareything, line 58: why not to use these on the connect side too?

scareything (approved):

inc_internal/message.h
ekoby: eventually they'll make it there

PR #778, opened 2024-11-27
CLI11 instead of subcommand

ekoby:


library/ziti.c
```diff
@@ -649,10 +650,14 @@ static void free_ztx(uv_handle_t *h) {
     FREE(ztx->session_token);
 
     ziti_ctrl_close(ztx_get_controller(ztx));
-    ztx->tlsCtx->free_ctx(ztx->tlsCtx);
+    if (ztx->tlsCtx) ztx->tlsCtx->free_ctx(ztx->tlsCtx);
```
dovholuknf, line 653: same with these changes. seems like a good thing just making sure they are intentional here

dovholuknf (approved):

PR #794, opened 2024-12-16
update ctrl list with new advertised address

ekoby:
better handle invalid controller responses

### DIFF
--- a/library/ziti.c
+++ b/library/ziti.c
@@ -537,8 +537,18 @@ static void on_ctrl_list_change(ziti_context ztx, const model_map *endpoints) {
 static void on_ctrl_redirect(const char *new_addr, void *ctx) {
     ziti_context ztx = ctx;
 
+    model_list_iter it = model_list_iterator(&ztx->config.controllers);
+    while(it) {
+        char *addr = (char*)model_list_it_element(it);
+        if (strcmp(addr, ztx->config.controller_url) == 0) {
+            it = model_list_it_remove(it);
+            free(addr);
+        }
+    }
     FREE(ztx->config.controller_url);
     ztx->config.controller_url = strdup(new_addr);
+    model_list_append(&ztx->config.controllers, strdup(new_addr));
+    
     ztx_config_update(ztx);
 }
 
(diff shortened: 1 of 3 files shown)


library/ziti.c
```diff
@@ -537,8 +537,18 @@ static void on_ctrl_list_change(ziti_context ztx, const model_map *endpoints) {
 static void on_ctrl_redirect(const char *new_addr, void *ctx) {
     ziti_context ztx = ctx;
 
+    model_list_iter it = model_list_iterator(&ztx->config.controllers);
+    while(it) {
+        char *addr = (char*)model_list_it_element(it);
+        if (strcmp(addr, ztx->config.controller_url) == 0) {
```
dovholuknf, line 543: probably should be case insensitive?

library/ziti_ctrl.c
```diff
@@ -237,9 +241,24 @@ static void ctrl_default_cb(void *s, const ziti_error *e, struct ctrl_resp *resp
     if (resp->new_address && strcmp(resp->new_address, ctrl->url) != 0) {
         CTRL_LOG(INFO, "controller supplied new address[%s]", resp->new_address);
 
+        const char *k;
+        ziti_controller_detail *detail;
+        MODEL_MAP_FOREACH(k, detail, &ctrl->endpoints) {
+            if (strcmp(k, ctrl->url) == 0) {
```
dovholuknf, line 247: same - case insensitive?

library/ziti_ctrl.c
```diff
@@ -237,9 +241,24 @@ static void ctrl_default_cb(void *s, const ziti_error *e, struct ctrl_resp *resp
...
+                break;
+            }
+        }
         FREE(ctrl->url);
         ctrl->url = resp->new_address;
         resp->new_address = NULL;
+        if(detail == NULL) {
+            detail = alloc_ziti_controller_detail();
+        }
+        FREE(detail->name);
```
dovholuknf, line 258: if detail was NULL previous, you still need to FREE  the name? seems like could be in the else?

dovholuknf (approved):
case insensitive comparisons seems like a good idea for a domain name?

PR #803, opened 2025-01-08
Avoid bind crypto errors

ekoby:
Need to preserve e2ee keys between bind since network may reuse existing terminator without updating key data

### DIFF
--- a/library/bind.c
+++ b/library/bind.c
@@ -26,18 +26,24 @@
 
 #define DEFAULT_MAX_BINDINGS 3
 #define REBIND_DELAY 1000
-#define REFRESH_DELAY (60 * 5 * 1000)
 
 #define CONN_LOG(lvl, fmt, ...) \
 ZITI_LOG(lvl, "server[%u.%u](%s) " fmt, \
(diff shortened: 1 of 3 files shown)


library/bind.c
```diff
@@ -110,21 +116,25 @@ static struct binding_s* new_binding(struct ziti_conn *conn) {
...
     struct ziti_ctx *ztx = conn->ziti_ctx;
 
-    size_t target = MIN(conn->server.max_bindings, model_map_size(&ztx->channels));
+    size_t target = MIN(conn->server.max_bindings,
+                        model_list_size(&conn->server.routers));
 
-    for(int idx = 0; conn->server.routers && conn->server.routers[idx]; idx++) {
-        ziti_edge_router *er = conn->server.routers[idx];
+    ziti_edge_router *er;
+    MODEL_LIST_FOREACH(er, conn->server.routers) {
```
scareything, line 137: list iterators work on model arrays?

scareything (approved):

library/bind.c
ekoby: it was changed to a list

PR #809, opened 2025-01-15
bump oidc callback buffer, log access token content

ekoby:


library/oidc.c
```diff
@@ -816,6 +816,25 @@ int oidc_client_close(oidc_client_t *clt, oidc_close_cb cb) {
     return 0;
 }
 
+static const char *jwt_payload(const char *jwt) {
+    static uint8_t payload[4096];
```
dovholuknf, line 820: imo, seems like it would be better to pass the length or use a `#define` ?

library/oidc.c
```diff
@@ -825,7 +844,12 @@ static void oidc_client_set_tokens(oidc_client_t *clt, json_object *tok_json) {
     if (clt->token_cb) {
         struct json_object *access_token = json_object_object_get(clt->tokens, "access_token");
         if (access_token) {
-            clt->token_cb(clt, OIDC_TOKEN_OK, json_object_get_string(access_token));
+            const char *token = json_object_get_string(access_token);
+            ZITI_LOG(DEBUG, "access_token=%s", jwt_payload(token));
```
dovholuknf, line 848: this will be really useful to debug! thx

library/oidc.c
```diff
@@ -502,7 +502,7 @@ static void ext_accept(uv_work_t *wr) {
         return;
     }
 
-    char buf[1024];
+    char buf[4096];
```
dovholuknf, line 505: some tokens are bigger than others... :)

dovholuknf (approved):

library/oidc.c
ekoby: this does not get receive token

PR #811, opened 2025-01-17
Allow better dev debugging

dovholuknf:
if ZITI_DEBUG is defined, change the log len to much longer and allow debugging the http responses from the controller

### DIFF
--- a/library/ziti_ctrl.c
+++ b/library/ziti_ctrl.c
@@ -430,6 +430,9 @@ static void ctrl_body_cb(tlsuv_http_req_t *req, char *b, ssize_t len) {
     if (len > 0) {
         if (resp->resp_content == ctrl_content_json) {
             if (resp->content == NULL) {
+#ifdef ZITI_DEBUG
+                CTRL_LOG(DEBUG, "HTTP RESPONSE: %s", b);
+#endif
                 resp->content = json_tokener_parse_ex(resp->content_proc, b, (int) len);
                 if (resp->content == NULL && json_tokener_get_error(resp->content_proc) != json_tokener_continue) {
                     CTRL_LOG(WARN, "parsing error: %s",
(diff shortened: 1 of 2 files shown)


library/ziti_ctrl.c
```diff
@@ -430,6 +430,9 @@ static void ctrl_body_cb(tlsuv_http_req_t *req, char *b, ssize_t len) {
     if (len > 0) {
         if (resp->resp_content == ctrl_content_json) {
             if (resp->content == NULL) {
+#ifdef ZITI_DEBUG
+                CTRL_LOG(DEBUG, "HTTP RESPONSE: %s", b);
+#endif
```
ekoby, line 435: you don't need this guard, and log at VERBOSE level
also `b` is not guaranteed to be `\0` terminated
you need to use `%.*s` formatting

library/ziti_ctrl.c
```diff
@@ -430,6 +430,9 @@ static void ctrl_body_cb(tlsuv_http_req_t *req, char *b, ssize_t len) {
     if (len > 0) {
         if (resp->resp_content == ctrl_content_json) {
             if (resp->content == NULL) {
+#ifdef ZITI_DEBUG
+                CTRL_LOG(DEBUG, "HTTP RESPONSE: %s", b);
```
scareything, line 434: should this specific log message require an opt-in of some kind (e.g. env variable)?

dovholuknf: i thought that making someone compile it specifically with this option enabled was "opt-in enough"?

ekoby: in a sense that was my suggestion -- you opt-in by setting the log level VERBOSE or TRACE are more appropriate

dovholuknf: removed the #define, changed to debug, and used `%.*s`

ekoby (approved):

PR #813, opened 2025-01-23
fix an issue with OIDC endpoints using different hostnames

ekoby:


library/oidc.c
```diff
@@ -708,26 +745,29 @@ int oidc_client_start(oidc_client_t *clt, oidc_token_cb cb) {
...
             {"audience",              clt->signer_cfg.audience ?
                                       clt->signer_cfg.audience : "openziti"},
     };
 
     int rc = 0;
     if (clt->mode == oidc_external) {
-        struct json_object *cfg = (struct json_object *) clt->config;
-        struct json_object *auth = json_object_object_get(cfg, "authorization_endpoint");
-        start_ext_auth(req, json_object_get_string(auth), sizeof(query)/sizeof(query[0]), query);
+        start_ext_auth(req, auth_url, sizeof(query)/sizeof(query[0]), query);
```
dovholuknf, line 762: `query[0]` will always be > 0?

dovholuknf (approved):
i feel like guarding against divide by zero would be a good idea but other than that seems good to me

library/oidc.c
ekoby: `sizeof(query[0])` is always > 0

PR #834, opened 2025-03-09
closes #833 - ensure code is urldecoded

### LINKED ISSUE #833: url encoded auth codes sent as form data
Google is returning a slash in the code returned after a successful OIDC flow. For example:
```
code=4%2F0AQSTgQHYQ
```

This code is then used as is by the c sdk to request the token however the code is url encoded and sent as url encoded. It needs to be url decoded before being used a form data

dovholuknf:
closes #833 - decode code passed in. some IdPs use codes with characters that need to be url encoded such as containing a `/` (`%2F`)

### DIFF
--- a/library/oidc.c
+++ b/library/oidc.c
@@ -270,7 +270,7 @@ static void internal_config_cb(oidc_req *req, int status, json_object *resp) {
             status = UV_EINVAL;
         }
     }
-    
+
     if (status == 0) {
         json_object_put(clt->config);
         clt->config = resp;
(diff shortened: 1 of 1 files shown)


library/oidc.c
```diff
@@ -507,6 +507,22 @@ static int set_blocking(uv_os_sock_t sock) {
     return 0;
 }
 
+void url_decode(const char *src, const char *end, char *dest, size_t dest_size) {
```
scareything, line 510: might as well make this static if we don't want external callers using it.

scareything (approved):

library/oidc.c
dovholuknf: done

PR #853, opened 2025-04-16
Ziti accept v2

ekoby:
Utilizes improved accept flow:
- reduce the number of control messages
- avoids race conditions in ER


plorenz (approved):
Looks good to me from a protocol perspective

scareything (approved):

PR #863, opened 2025-05-07
handle cert extension if requested by the network

ekoby:


qrkourier (approved):

library/ziti.c
```diff
@@ -2137,28 +2137,38 @@ static void api_session_cb(ziti_api_session *api_sess, const ziti_error *err, vo
...
 
-        struct tm exp;
-        ztx->id_creds.cert->get_expiration(ztx->id_creds.cert, &exp);
-        time_t now = time(0);
-        time_t exptime = mktime(&exp);
-
-        bool renew = exptime - now < ztx->opts.cert_extension_window * ONE_DAY;
-        if (!renew) {
+        if (!api_sess->is_cert_extendable) {
+            ZTX_LOG(DEBUG, "identity certificate is not renewable");
```
dovholuknf, line 2141: this feels like a warn to me

ekoby: primary cause of non-renewable certs is 3rd party CA identities. we don't want that noise at WARN level

library/ziti.c
```diff
@@ -2137,28 +2137,38 @@ static void api_session_cb(ziti_api_session *api_sess, const ziti_error *err, vo
...
+        struct tm exp;
+        if (api_sess->cert_extend_requested || api_sess->key_roll_requested) {
+            ZTX_LOG(INFO, "controller requested certificate renewal (%s key roll)",
+                    api_sess->key_roll_requested ? "with" : "without");
+        } else {
+            ztx->id_creds.cert->get_expiration(ztx->id_creds.cert, &exp);
+            time_t now = time(0);
+            time_t exptime = mktime(&exp);
+
+            bool renew = exptime - now < ztx->opts.cert_extension_window * ONE_DAY;
```
dovholuknf, line 2154: i know it used to be one day but one day really doesn't feel like enough imo unless when it comes back online it's allowed to immediately extend

ekoby: it isn't one day?

library/ziti.c
```diff
@@ -2137,28 +2137,38 @@ static void api_session_cb(ziti_api_session *api_sess, const ziti_error *err, vo
...
-            ZTX_LOG(WARN, "identity certificate needs to be renewed but application is not handling ZitiConfigEvent");
+            ZTX_LOG(WARN, "identity certificate needs to be renewed "
+                          "but application is not handling ZitiConfigEvent");
             goto done;
         }
 
-        ZTX_LOG(INFO, "renewing identity certificate exp[%04d-%02d-%02d %02d:%02d]",
-                1900 + exp.tm_year, exp.tm_mon + 1, exp.tm_mday, exp.tm_hour, exp.tm_min);
+        if (api_sess->key_roll_requested) {
+            ZTX_LOG(WARN, "key roll requested, but not yet supported");
```
dovholuknf, line 2169: seems like it should be up in line 2146

library/ziti.c
```diff
@@ -2137,28 +2137,38 @@ static void api_session_cb(ziti_api_session *api_sess, const ziti_error *err, vo
...
+        }
+
```
dovholuknf, line 2171: from a debugging perspective, i think there should be a marker log before this section and before the 'done' section below. It wasn't obvious to me that just getting here ends up with a new cert.

library/ziti.c
dovholuknf: seems like the sorta thing we could test for but i get ya. thx

library/ziti.c
ekoby: it is more of a TODO. this shouldn't be here long

PR #864, opened 2025-05-12
enhance json-c package finding

qrkourier:


library/CMakeLists.txt
```diff
@@ -47,7 +47,20 @@ if (ziti_DEVELOPER_MODE)
     endif ()
 endif ()
 
-find_package(json-c CONFIG REQUIRED)
+find_package(json-c CONFIG QUIET)
+if(NOT json-c_FOUND)
+    find_package(PkgConfig REQUIRED)
```
ekoby, line 52: - PkgConfig is already loaded here
- pkg_check_modules already creates a target (see how it loads `libsodium`). make it similar to that

PR #870, opened 2025-05-21
Misc improvements

ekoby:


library/CMakeLists.txt
```diff
@@ -124,6 +124,8 @@ function(config_ziti_library target)
     set_target_properties(${target} PROPERTIES
             C_STANDARD 11
             POSITION_INDEPENDENT_CODE ON
+            C_VISIBILITY_PRESET hidden
```
dovholuknf, line 127: these seem like they might have downstream impacts ... probably for the best, sure but it seems a little bit worrying?

dovholuknf (approved):

PR #875, opened 2025-05-22
try to get complete chain if controller indicates an issue with current certificates

ekoby:


library/ziti.c
```diff
@@ -2173,21 +2173,29 @@ static void api_session_cb(ziti_api_session *api_sess, const ziti_error *err, vo
...
+
+        if (api_sess->is_cert_improper) {
+            ZTX_LOG(INFO, "controller reported certificate chain as incomplete");
+            goto extend;
+        }
+
+        // check if identity cert is expiring or expired
+        if (ztx->opts.cert_extension_window > 0) {
+            if (!api_sess->is_cert_extendable) {
+                ZTX_LOG(DEBUG, "identity certificate is not renewable");
```
dovholuknf, line 2194: seems more like a WARN to me

dovholuknf (approved):

library/ziti.c
ekoby: i know you said it before, but it means it is a 3rd party cert, why warn about it?

dovholuknf: LOL i thought it felt familiar... there's no way to discern if it's from the controller's pki or 3rd party?

scareything (approved):

PR #914, opened 2025-09-18
Add Ziti_load_context_with_timeout

ffaraone:
Summary
Adds a new function to load an OpenZiti context with a timeout option. Unlike the existing load function, this version does not retry indefinitely when the OpenZiti network is unreachable. Instead, it fails gracefully after the specified timeout.

Motivation
When using the Python bindings on top of this C SDK, the current load behavior can cause the application to hang indefinitely if the network cannot be reached. This makes it difficult to handle error scenarios cleanly. Introducing a timeout provides predictable behavior and allows higher-level code to recover.

Details

Introduces a new function instead of modifying the existing one, ensuring backward compatibility.

The new function stops retrying after the timeout and returns control to the caller.

No changes to existing APIs or behavior for users who do not adopt the new function.

Next Steps
If accepted, I will open a corresponding PR in the Python SDK to expose this functionality.


antoniodimariano (approved):
OK

ekoby (commented):
please see comments

ffaraone:
Fixed!

ekoby (approved):
looks good

PR #969, opened 2025-12-15
implement stickiness

ekoby:


includes/ziti/ziti.h
```diff
@@ -283,7 +283,14 @@ typedef struct ziti_dial_opts_s {
...
     int connect_timeout_seconds;
-    char *identity;
+    const char *identity;
+    /**
+     * indicates that this connection is part of a group.
+     * grouped connections will use the same terminator if the service allows it (`service.routingStrategy == "sticky"`)
+     * this is useful when service can be more efficient when a client hits the same instance
+     * (client session caching for example)
+     */
+    const char *group;
```
scareything: I think calling this `terminator_group` would add a little clarity at the public api level.

scareything (approved):

includes/ziti/ziti.h
scareything: `terminator_label`? `terminator_tag`? I don't feel strongly. Maybe my brain is fusing the l2 feature into this but my first impression of "grouped connections" made me think of a bridge for multiple ziti connections.

PR #989, opened 2026-01-30
legacy auth: make sure a single refresh is active at any time

ekoby:


library/legacy_auth.c
dovholuknf: would it be possible to do the auth-refrehsing = false here instead of in the callbacks? that way on the off change we add another func here it's covered and if someone mistakenly removes that line it won't be an issue? i obviously don't know if it's important elsewhere or not but it seems like it could fit in here if not relevant in one of those callbacks.. (and if it is well obviously disregard lol)

ekoby: nope. the whole point is to prevent scheduling another refresh while HTTP request is processing -- between this and callback return

PR #991, opened 2026-02-11
fix: OIDC token refresh recovery after network disruption

dmuensterer:
A brief network disruption permanently breaks the OIDC tlsuv_http_t handle. Subsequent refresh attempts retry on the same broken handle and fail indefinitely. With only a 30-second refresh margin, retries are exhausted before the token expires, leading to UNAUTHORIZED errors that tear down all channels and require a full restart.

Three changes:

1. Reset broken HTTP handle after consecutive failures Add a refresh_failures counter. After 3 consecutive connection failures (code < 0), call tlsuv_http_cancel_all() to close the transport and reset the handle to Disconnected state, allowing tlsuv's auto-reconnect to establish a fresh connection. Also fixes a bug where the if-chain in refresh_cb used non-exclusive conditions, causing UV_EOF to trigger both "restart auth" and "5s retry" simultaneously.

2. Add 15s connect timeout to OIDC client The OIDC HTTP client was missing tlsuv_http_connect_timeout(), inheriting the kernel default (~130s). A single timeout could consume the entire refresh window. Now matches the controller client's 15s timeout.

3. Use half-lifetime refresh margin instead of fixed 30s Change token refresh scheduling from (expires_in - 30) to (expires_in / 2). For a typical 1800s token, this gives 900s of retry time (~120 attempts) instead of 30s (~6 attempts).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@ include(cmake/variables.cmake)
 set(tlsuv_DIR "" CACHE FILEPATH "developer option: use local tlsuv checkout")
 # developer setting: tlsuv branch or tag to use (if tlsuv_DIR is unset)
 if (NOT tlsuv_VERSION)
-    set(tlsuv_VERSION "v0.40.10")
+    set(tlsuv_VERSION "v0.40.12")
 endif (NOT tlsuv_VERSION)
 
 message("project version: ${PROJECT_VERSION}")
--- a/includes/ziti/ziti.h
+++ b/includes/ziti/ziti.h
@@ -689,7 +689,45 @@ extern const ziti_service *ziti_service_for_addr_str(ziti_context ztx, ziti_prot
 ZITI_FUNC
 extern const ziti_service *ziti_service_for_addr(ziti_context ztx, ziti_protocol proto, const ziti_address *addr, int port);
 
+/**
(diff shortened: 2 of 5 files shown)


library/oidc.c
```diff
@@ -349,10 +356,18 @@ static void auth_cb(tlsuv_http_resp_t *http_resp, const char *err, json_object *
...
+            return;
+        }
+        char *p = uri.query ? strstr(uri.query, "authRequestID=") : NULL; // guard against missing query string
+        if (p == NULL) { // guard against missing authRequestID parameter
+            failed_auth_req(req, "missing authRequestID in redirect");
+            return;
+        }
         p += strlen("authRequestID=");
-        req->id = strdup(p);
+        char *end = strchr(p, '&'); // stop at next query param to avoid capturing trailing params
```
dmuensterer: If the URL looks e.g. like: ` /oidc/authorize?authRequestID=abc123&state=xyz`

Then `req->id `becomes `abc123&state=xyz` instead of just `abc123`. That polluted ID then gets used in:

`snprintf(path, sizeof(path), "/oidc/login/cert?id=%s", req->id);`

ekoby:
this PR was pulled and superseded by #1025

PR #1031, opened 2026-04-12
Support additional query parameters on ext-jwt-signer authorize URL

smilindave26:
This is to enable auth/enrollment (e.g., enrollToCert) flows to take device info into account along with user info and make available as claims for use in identity name and external id creation.

Notes:
RFC 6749 §3.1 (OAuth 2.0 authorization endpoint): "The authorization server MUST ignore unrecognized request parameters."

OpenID Connect Core §3.1.2.1: "The Authorization Server MUST ignore unrecognized request parameters."

### DIFF
--- a/library/ext_oidc.c
+++ b/library/ext_oidc.c
@@ -141,6 +141,18 @@ int ext_oidc_client_init(uv_loop_t *loop, ext_oidc_client_t *clt,
 
     snprintf(clt->name, sizeof(clt->name), "%s", cfg->name ? cfg->name : cfg->provider_url);
     OIDC_LOG(INFO, "initializing with provider[%s]", cfg->provider_url);
+
+    // free previous auth params if re-initializing
+    if (clt->auth_params) {
+        for (int i = 0; i < clt->auth_params_count; i++) {
+            free((void *)clt->auth_params[i].name);
+            free((void *)clt->auth_params[i].value);
+        }
+        free(clt->auth_params);
+        clt->auth_params = NULL;
+        clt->auth_params_count = 0;
+    }
+
     clt->config = NULL;
     clt->tokens = NULL;
     clt->token_cb = NULL;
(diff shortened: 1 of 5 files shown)


library/ext_oidc.c
```diff
@@ -649,9 +654,28 @@ int ext_oidc_client_start(ext_oidc_client_t *clt, ext_oidc_token_cb cb) {
             {"state",                 req->state},
             {"audience",              clt->signer_cfg.audience ?
                                       clt->signer_cfg.audience : "openziti"},
+            {"device_name",           env->hostname},
```
dovholuknf: I would rather see the function take a list of params instead. it should be the responsibility of the caller of `ext_oidc_client_start` to add 'extra' params like this

smilindave26: Ok. Reworked this - the SDK now provides a generic ziti_ext_auth_set_params(ztx, params, count) API that lets the caller pass arbitrary name/value pairs. ext_oidc_client_start appends them to the authorize URL alongside the standard OIDC params. No hardcoded envInfo in the SDK.                                                                         
                                                                                                                         
The tunneler (ziti-tunnel-sdk-c) will be updated to call ziti_ext_auth_set_params with envInfo fields before ziti_ext_auth() in a follow-on PR.

library/external_auth.c
```diff
@@ -127,6 +127,36 @@ static void ext_token_cb(ext_oidc_client_t *oidc, enum ext_oidc_status status, c
...
+
+    if (params == NULL || count <= 0) {
+        return ZITI_OK;
+    }
+
+    // copy params
+    ztx->ext_auth->auth_params = calloc(count, sizeof(tlsuv_http_pair));
+    ztx->ext_auth->auth_params_count = count;
+    for (int i = 0; i < count; i++) {
+        ztx->ext_auth->auth_params[i].name = strdup(params[i].name);
```
dovholuknf: should the name be checked too?

smilindave26: updated

dovholuknf (approved):

PR #1034, opened 2026-04-15
improve external OIDC refresh

ekoby:
add jitter to refresh delay
attempt to use refresh_token even if access_token is expired (if present)

### DIFF
--- a/library/ext_oidc.c
+++ b/library/ext_oidc.c
@@ -46,7 +46,7 @@
 #define auth_url_path "/auth/callback"
 #define cb_url(host,port,path) "http://" host ":" _str(port) path
 #define default_cb_url cb_url("localhost",auth_cb_port,auth_url_path)
-#define default_scope "openid"
+#define default_scope "openid offline_access"
 
 #define AUTH_EP "authorization_endpoint"
 #define TOKEN_EP "token_endpoint"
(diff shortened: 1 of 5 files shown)


library/ext_oidc.c
```diff
@@ -760,36 +761,85 @@ static void ext_oidc_client_set_tokens(ext_oidc_client_t *clt, json_object *tok_
...
+        clt->signer_cfg.target_token == ziti_target_token_id_token ? "id_token" : "access_token");
+    if (jt_for_exp) {
+        const char *at = json_object_get_string(jt_for_exp);
+        if (zt_jwt_parse(at, &clt->current) != 0) {
+            // fallback for non-JWT tokens: use expires_in
+            struct json_object *ttl = json_object_object_get(clt->tokens, "expires_in");
+            if (ttl) {
+                uv_timeval64_t now;
+                uv_gettimeofday(&now);
+                clt->current.expiration = now.tv_sec + json_object_get_int(ttl);
```
dovholuknf, line 776: need to free `ttl`?

dovholuknf: nevermind - i see it's not needed here now

dovholuknf (approved):

PR #1041, opened 2026-04-27
Ext auth callback page updates

smilindave26:


library/ext_oidc.c
```diff
@@ -265,9 +331,48 @@ int ext_oidc_client_configure(ext_oidc_client_t *clt, oidc_config_cb cb) {
     return 0;
 }
 
+static void send_callback_response(uv_os_sock_t sock, const char *body) {
+    if (sock == INVALID_SOCK) return;
+
+    string_buf_t resp_buf;
+    string_buf_init(&resp_buf);
+    string_buf_fmt(&resp_buf, HTTP_RESP_FMT, 200, "OK", strlen(body), body);
```
ekoby: there is no reason to append the body -- just write HTTP response header and then body

library/ext_oidc.c
```diff
@@ -99,27 +106,80 @@ typedef struct auth_req {
...
+        if (error) {
+            string_buf_append(&buf, "        <dt>Error</dt><dd>");
+            append_html_escaped(&buf, error);
+            string_buf_append(&buf, "</dd>\n");
+        }
+        if (jwt) {
+            // jwt_payload returns a pointer to a static buffer that any
+            // future jwt_payload() call would clobber - copy it so the
+            // pretty-print fallback path remains valid.
+            char *raw = strdup(jwt_payload(jwt));
```
ekoby: no need to `strdup` the payload since you're parsing it

library/oidc_auth.c
```diff
@@ -172,12 +172,14 @@ static void token_cb(oidc_client_t *oidc, enum oidc_status status, const void *d
                 auth->mfa_cb(auth->cb_ctx, ZITI_OK);
                 auth->mfa_cb = NULL;
                 break;
-            case OIDC_TOKEN_FAILED:
-                snprintf(err, sizeof(err), "failed to auth: %d", status);
+            case OIDC_TOKEN_FAILED: {
+                const char *reason = data ? (const char *) data : "unknown";
+                snprintf(err, sizeof(err), "%s", reason);
                 auth->cb(auth->cb_ctx, ZitiAuthStateUnauthenticated, &(ziti_error){
                         .err = status,
```
ekoby: `ziti_error.err` should be a `ZITI_xxx` error

library/ext_oidc.c
```diff
@@ -331,35 +331,37 @@ int ext_oidc_client_configure(ext_oidc_client_t *clt, oidc_config_cb cb) {
...
+    return 0;
+}
+
+static void send_callback_response(uv_os_sock_t sock, const char *body) {
+    if (sock == INVALID_SOCK) return;
+
+    char header[256];
+    size_t body_len = strlen(body);
+    int header_len = snprintf(header, sizeof(header), HTTP_RESP_HEADER_FMT,
+                              200, "OK", body_len);
```
ekoby: should the code be 401 if token was rejected?

library/ext_oidc.c
smilindave26: updated

library/ext_oidc.c
smilindave26: gone

library/oidc_auth.c
smilindave26: yes

library/ext_oidc.c
smilindave26: yes, and 503 if interrupted

ekoby (approved):

PR #1066, opened 2026-06-11
feature: allow TOTP enrollment during internal OIDC flow

### LINKED ISSUE #1062: MFA enrollment is not possible when auth policy requires TOTP
```
[2026-06-03T19:59:04.363Z]    WARN ziti-sdk:ziti_ctrl.c:725 verify_api_session() ctrl[https://shawns-m1-mbp.localdomain:1280/edge/client/v1] no API session
[2026-06-03T19:59:04.363Z]   ERROR ziti-sdk:auth_queries.c:155 ziti_mfa_enroll_get_internal_cb() ztx[2] error during enroll MFA call: 0 - UNAUTHORIZED - no api session token set for ziti_controller
```

I think it's because `ziti_ctrl_get_mfa` first verifies the context has a, but there is no session because we haven't submitted an mfa token yet.

ekoby:
[fixes #1062]
[fixes #1017]

### DIFF
--- a/inc_internal/internal_model.h
+++ b/inc_internal/internal_model.h
@@ -175,7 +175,8 @@ XX(woken, model_bool, none, woken, __VA_ARGS__)
 #define ZITI_AUTH_QUERY_MFA_MODEL(XX, ...) \
 XX(id, model_string, none, id, __VA_ARGS__) \
 XX(type_id, ziti_auth_query_type, none, typeId, __VA_ARGS__) \
-XX(provider, model_string, none, provider, __VA_ARGS__) \
+XX(provider, model_string, none, provider, __VA_ARGS__)      \
+XX(enrolled, model_bool, none, isTotpEnrolled, __VA_ARGS__)  \
 XX(http_method, model_string, none, httpMethod, __VA_ARGS__) \
 XX(http_url, model_string, none, httpUrl, __VA_ARGS__) \
 XX(min_length, model_number, none, minLength, __VA_ARGS__) \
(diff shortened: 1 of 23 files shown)


includes/ziti/ziti_events.h
```diff
@@ -107,10 +107,12 @@ struct ziti_service_event {
 
 enum ziti_auth_action {
     ziti_auth_cannot_continue,
+    ziti_auth_enroll_totp,
     ziti_auth_prompt_totp,
```
scareything, line 111: existing enum values are changed. I think this deserves a `major` update?

scareything (approved):

inc_internal/internal_model.h
```diff
@@ -175,7 +175,8 @@ XX(woken, model_bool, none, woken, __VA_ARGS__)
 #define ZITI_AUTH_QUERY_MFA_MODEL(XX, ...) \
 XX(id, model_string, none, id, __VA_ARGS__) \
 XX(type_id, ziti_auth_query_type, none, typeId, __VA_ARGS__) \
-XX(provider, model_string, none, provider, __VA_ARGS__) \
+XX(provider, model_string, none, provider, __VA_ARGS__)      \
```
dovholuknf, line 178: from a down-stream consumer perspective like c-sharp, these sorts of changes being inserted into the struct change the byte offsets of every field after it. that and alignment/packing can make these sort of changes be problematic.

the same is true (as shawn points out below) with ziti_auth_enroll_totp. That can screw up csharp too

ekoby: this is internal model. C-sharp SDK should not use/access this

PR #1075, opened 2026-06-22
remove `offline_access` from default scopes

### LINKED ISSUE #1074: OIDC regression - `offline_access` scope added to defaults
In jan 2025 we [removed `offline_access` from the default scope list as too many IdPs don't support it](https://github.com/openziti/ziti-sdk-c/pull/806/changes)

[Two months ago we reintroduced it](https://github.com/openziti/ziti-sdk-c/commit/f66be3f52e8918d3d92e785ad8bc03ccc740c4f4) as a default scope which is a mistake. The only default scope that might make sense is `openid`. 

This has caused a problem with the latest mac client for users who have an IdP that doesn't support `offline_access`

dovholuknf:
closes #1074 - openid as the only default scope

### DIFF
--- a/library/ext_oidc.c
+++ b/library/ext_oidc.c
@@ -50,7 +50,11 @@
 #define auth_url_path "/auth/callback"
 #define cb_url(host,port,path) "http://" host ":" _str(port) path
 #define default_cb_url cb_url("localhost",auth_cb_port,auth_url_path)
-#define default_scope "openid offline_access"
+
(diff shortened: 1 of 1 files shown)


ekoby:
`offline_access` is useful if available, this should check if IdP allows that (from `/.well-known/openid-configuration`.`scopes_supported`) and add it if it is

dovholuknf:
the external signer has a place for scopes. the administrator should be able to choose which scopes they want to support not the csdk

ekoby (approved):

Issue #990, opened 2026-02-10
OIDC token refresh never recovers from network disruption, permanent UNAUTHORIZED state requires restart


dmuensterer:
# OIDC token refresh never recovers from brief network disruption — permanent UNAUTHORIZED state requires restart

## Summary

A brief network disruption (subsecond TCP RST) permanently kills OIDC-authenticated `ziti-edge-tunnel` clients. The OIDC token refresh loop retries on a broken HTTP handle that never reconnects, exhausts a 30-second retry window, and the client enters permanent `UNAUTHORIZED` state. Only a manual process restart recovers it.

This currently blocks us from upgrading from legacy authentication to OIDC. We have been unable to roll out OIDC auth because in practice it means tunnelers permanently drop roughly once a day, requiring manual intervention or external watchdog restarts.

## Environment

- **ziti-edge-tunnel** v1.10.9
- Single-controller deployment (non-HA)
- OIDC authentication

## What happens

After any event that kills an established TCP connection to the controller (network flap, LB idle timeout, firewall state flush etc.), the timeline is:

1. Both the controller API and OIDC HTTP handles discover the dead connection on their next use
2. OIDC refresh retries every 5s on the same broken `tlsuv_http_t` handle, every attempt fails in ~30ms with `-103`/`-9`
3. After 30 seconds the access token expires (hardcoded as per `library/oidc.c:567-580`)
4. `ztx_prepare()` triggers `ziti_set_unauthenticated()` → all edge router channels torn down → all tunneled traffic severed
5. Recovery attempts (`ziti_force_api_session_refresh`, `ziti_re_auth`) go through the same broken handles → same instant failures
6. Client is permanently dead

With a 30-minute token, any disruption has a ~1.7% chance of overlapping the final 30-second refresh window. Over a day's worth of tokens (48 cycles), the probability of at least one hit is ~56%. In practice this means **almost daily permanent failures** in our environment with occasional network imperfections.

## Log evidence

```
# Connection dies, controller API fails instantly (30ms, not a fresh TCP attempt):
(3716046)[     1946.837] ctrl_resp_cb() request[/edge/client/v1/current-api-session] failed: -103(software caused connection abort)
(3716046)[     1946.866] ctrl_resp_cb() request[/edge/client/v1/current-api-session] failed: -9(bad file descriptor)

# OIDC refresh hits the same broken handle, retries every 5s, never recovers:
[...]

Issue #572, opened 2023-11-13
unable to recover API session


qrkourier:
I'm running a Python edge SDK app continuously, which uses the C SDK, and it became de-authorized a few hours ago and seems to be stuck, unable to recover. I was able to curl `/edge/client/v1/version` while the edge SDK was reporting "CONTROLLER_UNAVAILABLE" so that message appears to be inaccurate:

```
2023-11-13T13:48:47.920Z     WARN ziti-sdk:ziti_ctrl.c:487 verify_api_session() ctrl[db2f5d29-d062-47e1-82dc-a12738f0768e.production.netfoundry.io] no API session
2023-11-13T13:48:47.920Z     WARN ziti-sdk:bind.c:212 session_cb() server[0.0] failed to get session for service[classifier-service]: -14/UNAUTHORIZED
2023-11-13T13:48:52.655Z     INFO ziti-sdk:ziti.c:913 ziti_re_auth_with_cb() ztx[0] starting to re-auth with ctlr[https://db2f5d29-d062-47e1-82dc-a12738f0768e.production.netfoundry.io:443] api_session_status[0] api_session_expired[TRUE]
2023-11-13T13:48:57.759Z    ERROR ziti-sdk:ziti_ctrl.c:162 ctrl_resp_cb() ctrl[db2f5d29-d062-47e1-82dc-a12738f0768e.production.netfoundry.io] request failed: -103(software caused connection abort)
2023-11-13T13:48:57.759Z     WARN ziti-sdk:ziti.c:1597 api_session_cb() ztx[0] failed to get api session from ctrl[https://db2f5d29-d062-47e1-82dc-a12738f0768e.production.netfoundry.io:443] api_session_state[1] CONTROLLER_UNAVAILABLE[-16] software caused connection abort
```

To recover, I bounced the Python application. It started normally.

This is the beginning of the interruption event time frame from the application log, ~5 hours earlier.

```
2023-11-13T09:07:29.789Z     INFO ziti-sdk:ziti.c:1560 ziti_set_api_session() ztx[0] api session set, setting api_session_timer to 1740s
2023-11-13T09:36:18.455Z    ERROR ziti-sdk:channel.c:831 on_channel_data() ch[0] unhandled error on_data [-110/connection timed out]
2023-11-13T09:36:18.455Z     WARN ziti-sdk:bind.c:335 on_message() binding failed: -21/connection to edge router terminated
[...]

PR #1086, opened 2026-07-11
fix: recover from a stuck legacy (cert-based) auth/refresh request

bjornmage:
## Symptom

A cert-based (non-OIDC) identity lost its API session and never recovered on
its own -- it looped on `UNAUTHORIZED/no api session token set for
ziti_controller` continuously for hours, well past the point the underlying
disruption had cleared, until the process was manually restarted. This is the
same defect class as #990 ("OIDC token refresh never recovers from network
disruption, permanent UNAUTHORIZED state requires restart"), but #990's fix
(#1025) is scoped entirely to `oidc.c`/`oidc.h`/`credentials.c` -- the OIDC
token-refresh path. The identity in this report uses cert-based auth
(`--identity <file>.json`, not OIDC enrollment), so #1025's fix does not
cover this code path. #572 (2023, general controller-API "stuck, never
reconnects" pattern, not OIDC-specific, closed without an evident code fix)
looks like the same still-open gap in the non-OIDC path.

Environment: `ziti-edge-tunnel`/`ziti-sdk-c` 1.18.1, `tlsuv` v0.41.4, 3-node
Raft HA controller cluster, 30-minute API session timeout. Client process
never crashed or exited (`systemctl` showed `active (running)` throughout),
so nothing triggered `Restart=always`; only an explicit process restart
recovered it, and the fresh process authenticated immediately.

```
# Steady-state failure loop, repeating every ~10s for hours:
ERROR ziti-sdk:ziti.c:1728 update_identity_data() ztx[1] failed to get identity_data: no api session token set for ziti_controller[UNAUTHORIZED]
ERROR ziti-sdk:ziti.c:1647 edge_routers_cb() ztx[1] failed to get current edge routers: code[0] UNAUTHORIZED/no api session token set for ziti_controller
ERROR ziti-sdk:connect.c:524 process_connect() conn[...] ziti context is not authenticated, cannot connect to service[...]
ERROR tunnel-cbs:ziti_tunnel_cbs.c:128 on_ziti_connect() ziti dial failed: invalid state
```

## Root cause

`tlsuv_http` exposes a connect timeout (`tlsuv_http_connect_timeout()`) but
no idle/response timeout. In `legacy_auth.c`, once a refresh/auth request
completes its TCP handshake, there is no bound on how long it can wait for a
response. If the peer accepts the connection and then never responds (e.g. a
controller instance that disappears mid-request during a rolling restart,
which is what coincided with the session invalidation in this report),
`legacy_session_cb()` is simply never invoked. `auth->refreshing` stays
latched `true` forever, and the existing retry/backoff logic in
`legacy_session_cb()` -- which does work correctly for every other failure
mode -- never gets a chance to run, because it's never reached.

## Fix

Add a bounded watchdog timer (`AUTH_REQUEST_TIMEOUT_SECONDS`, 30s) around
each outstanding legacy refresh/auth request (`library/legacy_auth.c`). On
expiry it calls `tlsuv_http_cancel_all()` on the auth context's HTTP client,
which routes the stuck request back through `legacy_session_cb()` with
`UV_ECANCELED` -- the same transport-failure path already exercised for
other network errors, and it re-arms the existing retry/backoff calculation
unchanged. The watchdog timer is started alongside every outstanding
request and stopped as soon as a response (successful or not) arrives, so it
never fires for healthy sessions.

This mirrors the recovery semantics #1025 added for the OIDC auth path,
applied to the legacy/cert-based path that #1025 did not cover.

Handle lifecycle: `req_timer` is closed and its close callback chains into
closing the existing `timer` handle (whose close callback frees `auth`), so
`auth` is only freed after both embedded `uv_timer_t` handles have finished
closing.

No behavior change for healthy sessions: the watchdog only fires when a
request has gone unanswered past `AUTH_REQUEST_TIMEOUT_SECONDS`; the
existing delay/backoff calculations (`refresh_delay()`, `next_backoff()`)
are untouched.

## Test evidence

- Full existing unit test suite (`ctest`, Catch2, 83 tests / ~100k
  assertions) passes unchanged against the patched build -- no regressions.
- Added a standalone regression harness (not part of this PR, used for local
  verification) that opens a mock controller which accepts a connection and
  then goes silent, confirming: the watchdog fires at ~30s, cancels the
  stuck request, the existing backoff path schedules a retry, the retry
  opens a fresh connection, and a subsequent valid response drives the auth
  state machine to `ZitiAuthStateFullyAuthenticated`.
- Not verified: end-to-end recovery against a live controller cluster
  reproducing the exact rolling-restart timing from the original incident
  (`tests/integ` requires a live quickstart Ziti network and was out of
  scope for this change's local verification).

## Related

- #990 -- same "permanent UNAUTHORIZED, restart required" symptom, fixed by
  #1025, but scoped to OIDC auth only.
- #1025 -- the OIDC-path fix this PR mirrors for the legacy/cert-based path.
- #572 -- closed 2023, no evident code fix; same "stuck, controller API
  handle never reconnects" pattern via the general controller-API path,
  likely the same defect class as #990 in the non-OIDC path.

### DIFF
--- a/library/legacy_auth.c
+++ b/library/legacy_auth.c
@@ -23,6 +23,15 @@
 #define API_SESSION_DELAY_WINDOW_SECONDS 60
 #define API_SESSION_EXPIRATION_TOO_SMALL_SECONDS 120
 
+// tlsuv_http only exposes a connect timeout -- once a request is past connect
+// it can be left dangling indefinitely (e.g. a peer that accepts the TCP
+// handshake and then never responds, as seen when a controller instance
+// disappears mid-request during a rolling restart). Without a bound here,
+// a single stuck refresh/auth request latches `refreshing` permanently and
+// the auth state machine never retries again. See fix for permanent
+// UNAUTHORIZED after API session expiry, cert-based auth path.
+#define AUTH_REQUEST_TIMEOUT_SECONDS 30
+
 #define AUTH_LOG(lvl, fmt, ...) ZITI_LOG(lvl, "legacy_auth[%s]: " fmt, auth->http.host, ##__VA_ARGS__)
 
 struct legacy_auth_s {
@@ -33,6 +42,7 @@ struct legacy_auth_s {
 
     tlsuv_http_t http;
     uv_timer_t timer;
+    uv_timer_t req_timer; // watchdog: bounds how long an in-flight refresh/auth request can go unanswered
     ziti_api_session *session;
     
     bool has_x509;
@@ -63,6 +73,7 @@ static void free_body_cb(tlsuv_http_req_t *req, char *body, ssize_t len) {
 }
 
 static void legacy_timer_cb(uv_timer_t *t);
+static void legacy_req_timeout_cb(uv_timer_t *t);
 
 #define LEGACY_AUTH_INIT()          \
     (ziti_auth_method_t) {          \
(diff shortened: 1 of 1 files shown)


ekoby:
recheck cla

PR #1092, opened 2026-07-17
Auth method transition

### LINKED ISSUE #1089: implement live auth method transition (legacy -> OIDC)
when network is upgraded from legacy (pre-2.0-no-OIDC) to 2.0+ ziti context should switch auth method to OIDC without requiring restart

ekoby:
Goal: 
when network is updated and gains internal OIDC authentication (automatic when updating to ziti 2.0+), ziti context should switch to it without requiring restart

Implementation:
`ziti_ctrl` automatically updates controller version whenever `ziti-instance-id` changes. controller capabilities are cached
ziti_context detects when OIDC auth method becomes available and replaces auth_method

Caveats:
User may need to complete secondary authentication steps (external auth, or TOTP) to gain full functionality

[fixes #1089]

### DIFF
--- a/library/ziti.c
+++ b/library/ziti.c
@@ -51,11 +51,11 @@
 #define ONE_DAY (60 * 60 * 24)
 
 #define ztx_controller(ztx) \
-((ztx)->ctrl.url ? (ztx)->ctrl.url : (ztx)->config.controller_url)
+(cstr_is_empty(&(ztx)->ctrl.url) ? cstr_str(&(ztx)->ctrl.url) : (ztx)->config.controller_url)
 
 int code_to_error(const char *code);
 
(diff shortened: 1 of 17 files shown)


library/ziti.c
```diff
@@ -1920,6 +1921,11 @@ void ztx_prepare(uv_prepare_t *prep) {
     ziti_context ztx = prep->data;
     ZTX_LOG(TRACE, "preparing ztx for IO");
 
+    if (ztx->auth_method && ztx->auth_method->kind == LEGACY) {
```
scareything, line 1924: I know it's not preferred, but do we need to worry about oidc --> legacy?

scareything (approved):

PR #1100, opened 2026-07-24
external auth improvements and fixes

ekoby:
- keep track of primary vs secondary ext auth (enable dual JWT login -- why would anyone..)
- fix auth with (cert-only + jwt-secondary) policies

### DIFF
--- a/library/ext_oidc.c
+++ b/library/ext_oidc.c
@@ -14,11 +14,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <oidc.h>
+#include "credentials.h"
+
 #include <assert.h>
-#include <json-c/json.h>
-#include <sodium.h>
 #include <ctype.h>
+#include <json-c/json.h>
+#include <oidc.h>
+#include <sodium.h>
 #ifdef _WIN32
 #include <winsock2.h>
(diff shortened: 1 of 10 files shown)


library/ext_oidc.c
```diff
@@ -296,11 +305,17 @@ static void internal_config_cb(tlsuv_http_resp_t *r, const char * err, json_obje
         clt->refresh_grant = "refresh_token";
         // config has full URLs, so we can drop the prefix now
         tlsuv_http_set_path_prefix(&clt->http, "");
+        json_object *iss = json_object_object_get(clt->config, "issuer");
+        if (iss == NULL) {
+            OIDC_LOG(ERROR, "issuer is not provided!!!");
```
dovholuknf, line 308: why would we allow this branch to continue on? no issuer is a big deal right? is it somehow defaulted elsewhere? i haven't quite figured out if this is or is not a big deal

ekoby: `/.well-known/openid-configuration.issuer` is required by the spec but nothing here relies on it. as long as correct issuer is put in JWTs it will work.

issuer field is only used for logging so we can continue even if it is missing

dovholuknf (approved):

PR #1112, opened 2026-08-07
tear down connections after consecutive transport errors

scareything:
I observed a stuck auth retry cycle with ziti-sdk-c 1.18.5. This change is currently untested. logs attached.

[zde.refresh.failed.zip](https://github.com/user-attachments/files/30835638/zde.refresh.failed.zip)


scareything:
doesn't help

PR #1121, opened 2026-08-24
ha enrollment: populate controller endpoints from response

ekoby:
populate identity file `ztAPIs` from the list returned in enrollment response

see openziti/ziti#3947

### DIFF
--- a/library/ziti_enroll.c
+++ b/library/ziti_enroll.c
@@ -450,6 +450,18 @@ static void enroll_cb(ziti_enrollment_resp *resp, const ziti_error *err, void *e
 
     ZITI_LOG(DEBUG, "successfully enrolled with controller %s", ziti_ctrl_get_url(&er->controller));
     er->cfg.id.cert = resp->cert ? strdup(resp->cert) : strdup(er->opts.cert);
+    if (model_list_size(&resp->controllers) > 0) {
+        model_list_clear(&er->cfg.controllers, free);
+        ziti_controller_detail *ctrl;
+        api_address *addr;
+        MODEL_LIST_FOREACH(ctrl, resp->controllers) {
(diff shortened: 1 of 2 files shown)


library/ziti_enroll.c
```diff
@@ -450,6 +450,18 @@ static void enroll_cb(ziti_enrollment_resp *resp, const ziti_error *err, void *e
...
     ZITI_LOG(DEBUG, "successfully enrolled with controller %s", ziti_ctrl_get_url(&er->controller));
     er->cfg.id.cert = resp->cert ? strdup(resp->cert) : strdup(er->opts.cert);
+    if (model_list_size(&resp->controllers) > 0) {
+        model_list_clear(&er->cfg.controllers, free);
+        ziti_controller_detail *ctrl;
+        api_address *addr;
+        MODEL_LIST_FOREACH(ctrl, resp->controllers) {
+            MODEL_LIST_FOREACH(addr, ctrl->apis.edge) {
+                if (addr->url != NULL) {
+                    model_list_append(&er->cfg.controllers, strdup(addr->url));
```
dovholuknf, line 460: i feel like this could be more robust, at very least just making sure the url is a legit would be a good first step..  probing the url to make sure it's actually available and warning when it's not would just be good but perhaps that is done somewhere else, there's probably other stuff to do as well to confirm it's valid seem like a good idea...

dovholuknf (approved):
i'd like to see the url validated/verified but i'll leave that to you

library/ziti_enroll.c
ekoby: not sure what you mean, the controller that reports that is verified with JWT signature, you have to trust whatever it tells you, right?

availability check is a bad idea because you when you try to load identity later the available controller set might be different (that's the whole reason for this change)

PR #1124, opened 2026-08-25
don't collect posture responses that haven't been answered yet

### LINKED ISSUE #1123: crash sending posture when process response hasn't been gathered yet
`ziti_collect_posture(..., true)` appends entries whose `obj` is still NULL (posture.c:592), then `create_posture_resp` derefs `obj->typeId` (posture.c:645). Any process posture check on a dialed service crashes. 1.18.2 and 1.18.7. sometime between 1.10.4 and 1.18 this happened but it ended up causing a crash in a test so it seems like it should be fixed.

dovholuknf:
closes #1123 

and align 'collect_posture' to the 'ziti_' convention

### DIFF
--- a/library/posture.c
+++ b/library/posture.c
@@ -576,7 +576,7 @@ static char *ziti_pr_to_json(const ziti_pr_base *pr) {
 
 int model_list_fmt_to_json(string_buf_t *buf, model_list *l, const type_meta *meta, int flags, int indent);
 
-void ziti_collect_posture(ziti_context ztx, model_list *send_prs, bool collect_all) {
+void ztx_collect_posture(ziti_context ztx, model_list *send_prs, bool collect_all) {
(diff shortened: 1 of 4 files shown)


library/posture.c
```diff
@@ -1292,7 +1293,7 @@ void ziti_send_posture_er(ziti_context ztx, ziti_channel_t *ch) {
     }
 
     model_list send_prs = {};
-    collect_posture(ztx, &send_prs, true);
+    ziti_collect_posture(ztx, &send_prs, true);
```
ekoby: internal functions should be `ztx_` (if first arg is `ziti_context`)

`ziti_` prefix is reserved for external API

dovholuknf: fixed

ekoby (approved):

PR #1125, opened 2026-08-25
fix posture test

dovholuknf:
capture the posture callback so the fixture never reaches the real send path


tests/posture_tests.cpp
```diff
@@ -76,7 +83,12 @@ namespace {
             ztx.posture_checks = nullptr;
             model_map_clear(&ztx.services, (_free_f) free_ziti_service_ptr);
             uv_loop_close(&loop);
```
ekoby: is loop even needed?

dovholuknf: dropped the init/close and commented it. only used for uv_now in ziti_posture_init's ztx_set_deadline.

vcpkg.json
```diff
@@ -11,6 +11,11 @@
       "version": "1.52.0",
       "port-version": 0,
       "$comment": "1.52.1 breaks on mingw so we're sticking with 1.52.0 until fix released and available in vcpkg"
+    },
+    {
```
ekoby, line 15: drop this and rebase

ekoby (changes requested):

ekoby (approved):

PR #1131, opened 2026-08-27
Guard getdomainname() on Android API < 26

NicFragale:
getdomainname() was not exposed by bionic libc until API level 26 (Android 8.0). ziti_info_init() calls it unconditionally on any non-Windows platform, which fails to link when targeting older Android API levels (e.g. API 22, needed for devices still running Android 5.x/Lollipop-era firmware).

There is no NIS-style domain name concept on Android, so on affected API levels the domain is simply left empty, matching the existing fallback behavior when getdomainname() itself fails at runtime.


NicFragale:
I have read the CLA Document and I hereby sign the CLA

ekoby:
recheck cla

ekoby (approved):

PR #1135, opened 2026-08-31
retry OIDC auth failure to allow HA cluster to catchup

### LINKED ISSUE #1134: 1.18.8+ HA: post-enrollment auth may choose a controller with unreplicated data, fails, never retries
if the csdk picks a node other than the one it just enrolled against, auth fails — that peer doesn't 
have the identity's authenticator yet — and the sdk never retries, leaving the identity disconnected
until something reloads it

On an HA cluster, an identity enrolled through AddIdentity comes up disconnected and stays that way. 
The tunneler enrolls against the leader and immediately loads the identity; the SDK then picks one of 
the controllers from theenrollment response at random and authenticates there. If that isn't the node 
it just enrolled against, POST /oidc/login/cert returns 401 INVALID_AUTH because that peer hasn't 
applied the authenticator yet. The SDK treats the rejection as final — no retry, no failover to the 
other controllers it is holding — so the identity is disconnected for the life of the process.

Observed in the ziti-tunnel-sdk-c nightly HA run (using ziti-sdk-c 1.18.8, controller v2.1.0, 3 nodes)

---

**Claude observations:**

- Cert-auth failures by node: leader 0, followers 4 and 1. Every follower rejection is followed ~25ms later by the raft entry that would have made it succeed.
- INVALID_AUTH appears 22× on 1.18.8 and 0× on 1.18.7.
- The whole failure is one unbroken millisecond: login_cb 401 INVALID_AUTH → failed_auth_req → ztx_set_unauthenticated ztx[41] → on_event ... failed to connect to controller → {"Op":"controller","Action":"disconnected"}.
- No retry: that context emitted ten log lines total while the tunneler kept running for another 56 seconds.
- Not MFA-specific — test_on_off and test_url_enroll_after_jwt show the identical chain.

Bisected to b6192852 (PR #1121): 1.18.7 passes, 1.18.7 + that one commit fails, 1.18.8 fails. Signature test TestMFAEnrollment/enrollCompletesWithTotpRequiredPolicy — 0.21s pass vs 5s timeout.

ekoby:
- retry window: 2 minutes with exponential backoff
- deleted identities will fail and report impossible to authenticate stated after the window

[fixes #1134]

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -21,6 +21,7 @@ add_executable(all_tests
         e2ee_tests.cpp
         util_tests.cpp
         oidc_recovery_tests.cpp
+        oidc_auth_retry_tests.cpp
         ctrl_endpoint_tests.cpp
         posture_tests.cpp)
 
--- a/inc_internal/oidc.h
+++ b/inc_internal/oidc.h
@@ -67,6 +67,13 @@ struct oidc_client_s {
     tlsuv_http_req_t *refresh_req;
     int refresh_failures;
 
+    // initial/full auth flow retry state: a freshly enrolled identity may not
+    // have replicated to the controller node we reached yet, so the login leg
+    // rejects good credentials for a short while. see schedule_auth_retry().
+    int auth_failures;            // consecutive auth-flow failures, feeds next_backoff()
+    uint64_t auth_retry_until;    // uv_now() ms past which a failure is reported; 0 = no window open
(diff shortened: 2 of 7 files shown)


tests/CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ add_executable(all_tests
         e2ee_tests.cpp
         util_tests.cpp
         oidc_recovery_tests.cpp
+        oidc_auth_retry_tests.cpp
```
dovholuknf, line 24: forget to commit this maybe?

dovholuknf (approved):
some tests perhaps to add sent offline. lgtm

PR #1137, opened 2026-09-01
macOS/iOS: use SecurityFramework for e2ee:AES-GCM mode

ekoby:


smilindave26 (approved):

library/e2ee/e2ee_aes_gcm_apple.c
```diff
@@ -0,0 +1,445 @@
...
+#include <string.h>
+
+#include <CoreFoundation/CoreFoundation.h>
+#include <Security/Security.h>
+#include <CommonCrypto/CommonCrypto.h>
+
+#include "crypto.h"
+#include "ziti/ziti_log.h"
+
+// Apple ships no public C API for raw AES-GCM or for HKDF: CryptoKit is Swift-only and
```
scareything, line 27: I think we should attempt to load these symbols with dlopen instead of declaring them here:
1. if the symbols are not present then we have a crash that will be difficult to diagnose.
2. the presence of these symbols in the binary is likely to get us flagged in an App Store review.

